PR #8, opened 2024-02-06
Add new user permissions

jsangmeister:
copied from https://github.com/OpenSlides/openslides-backend/pull/2158/


jsangmeister:
> @jsangmeister can you have a look and merge this PR? The autoupdate-service already has a merged PR based on these changes.

Yes, but then we'll have to merge the backend as well, and should do the same for the client

PR #17, opened 2024-02-07
generate relational schema

r-peschke:


models.yml
```diff
@@ -8,18 +8,25 @@
...
+#       in case of n:m-relations with an intermediary table. 
 #     - Non-generic relations: The simple syntax for such a field
-#       `to: <collection>/<field>`. This is a reference to a collection. The reverse
-#       relation field in this collection is <field>. E. g. in a motion the field `meeting_id`
-#       `reference: <collection(field1, ...)` This can be used, if the foreign key reference
-#        is not the <collection/"id">, where the collection is the same as that from the `to`.
+#       `to: <collection>/<field>`. This is a reference to a collection-field, the "other" side of a relation.
+#           In the relational DB this will be a field in a view filled by an automatically generated sql.
+#           During the development of the relational db one relation can have a `to` and a `reference` property.
+#           In this case the `to` value is only used for historical reason, onky the `reference` property is used
```
jsangmeister, line 23:
```suggestion
#           In this case the `to` value is only used for historical reason, only the `reference` property is used
```

models.yml
```diff
@@ -8,18 +8,25 @@
...
+#           for the schema generation.  
+#           If necessary the sql can be written manually, see `sql`.
+#       `reference: <collection(field1, ...)` This will create a foreign key constraint. Therefore a `reference` is
+#           the indicator for a real field in the table and you cannot use it for relation types with a `list` suffix.
+#           For some time there may be also a `to`-property for the relation, but for schema generation 
+#           the `reference` takes the prededence over the `to`.
```
jsangmeister, line 29:
```suggestion
#           the `reference` takes prededence over the `to`.
```

models.yml
```diff
@@ -35,6 +42,10 @@
 #           - motion/option_ids
 #           - user/option_ids
 #           - poll_candidate_list/option_id
+#       The `reference`, only for generic-relation, but not generic-relation-list will be a list of references:
```
jsangmeister, line 45:
```suggestion
#       For a generic-relation (not for generic-relation-list, see above), the attribute `reference` is a list of references:
```

models.yml
```diff
@@ -45,13 +56,18 @@
 #     - You can add JSON Schema properties to the fields like `enum`, `description`,
 #       `items`, `maxLength` and `minimum`
 # Additional properties:
+#     - The property `sql` can be filled manually with an sql-statement, which fills this field. This
+#       field can't be filled with an own vaue and is always defined in a view. You can use it on all
+#        field types, not only relations.
```
jsangmeister, line 61: Too many "fills" and "filled"... Maybe something like this is clearer:
```suggestion
#     - The optional property `sql` may contain an SQL statement which is used to calculate the value of this field. This
#       field does not exist with its own value in a table, but is always defined in a view. It can be used on all
#        field types, not only relations. 
```

jsangmeister (changes requested):
The concept itself looks good. I'm missing a list of required attributes for reach relation type, e.g.:
- 1:1 must always contain a `reference` on one side and a `sql` on the other
- similar rules for 1:n and m:n

(just an example because I don't know the actual rules yet)

This can then be validated if we want via the `to` attributes, which is another reason why I would like to keep them.

models.yml
```diff
@@ -8,18 +8,25 @@
...
-#       `reference: <collection(field1, ...)` This can be used, if the foreign key reference
-#        is not the <collection/"id">, where the collection is the same as that from the `to`.
+#       `to: <collection>/<field>`. This is a reference to a collection-field, the "other" side of a relation.
+#           In the relational DB this will be a field in a view filled by an automatically generated sql.
+#           During the development of the relational db one relation can have a `to` and a `reference` property.
+#           In this case the `to` value is only used for historical reason, onky the `reference` property is used
+#           for the schema generation.  
+#           If necessary the sql can be written manually, see `sql`.
+#       `reference: <collection(field1, ...)` This will create a foreign key constraint. Therefore a `reference` is
+#           the indicator for a real field in the table and you cannot use it for relation types with a `list` suffix.
```
jsangmeister, line 27: Also, this is not the case in many places throughout the `model.yml`

models.yml
```diff
@@ -8,18 +8,25 @@
...
-#       relation field in this collection is <field>. E. g. in a motion the field `meeting_id`
-#       `reference: <collection(field1, ...)` This can be used, if the foreign key reference
-#        is not the <collection/"id">, where the collection is the same as that from the `to`.
+#       `to: <collection>/<field>`. This is a reference to a collection-field, the "other" side of a relation.
+#           In the relational DB this will be a field in a view filled by an automatically generated sql.
+#           During the development of the relational db one relation can have a `to` and a `reference` property.
+#           In this case the `to` value is only used for historical reason, onky the `reference` property is used
+#           for the schema generation.  
+#           If necessary the sql can be written manually, see `sql`.
+#       `reference: <collection(field1, ...)` This will create a foreign key constraint. Therefore a `reference` is
```
jsangmeister, line 26: - fix the `<>` brackets (and use them either for all variables or for none)
- is it correct that it's possible to give multiple fiel names here? If so, how should that work? In this case, an explanation in the docs would be helpful, although I'm not sure if we even need this feature. If it's not possible or if you agree that we do not need this feature, please correct the description
- Currently, the `id` field is the only field that is ever used here. Since all models have an id and one normally only wants to reference models by their id, I don't see this changing anytime soon. Therefore, we could remove the need to append the `(id)` suffix for all references and simply write `reference: <collection>`. Referencing anything other than the id sounds like a bad design idea, anyway. What do you think?

models.yml
r-peschke: > Also, this is not the case in many places throughout the `model.yml`

Yes, I told you and this will change

.gitignore
```diff
@@ -32,3 +32,5 @@ secrets/
 
 # code-workspaces
 *.code-workspace
+
+.gitignore
```
jsangmeister, line 36: Ignoring the `.gitignore` cannot be right...

dev/src/generate_sql_schema.py
```diff
@@ -43,6 +43,12 @@ class SQL_Delete_Update_Options(str, Enum):
     NO_ACTION = "NO ACTION"
 
 
+class FIELD_SQL_ERROR_ENUM(Enum):
```
jsangmeister, line 46: I assume this differentiates between actual fields and calculated fields? Then I would prefer a more articulate name, e.g. `FieldType`. Listing the properties in the name is not a good naming pattern.

Also, you seem to use different casings for the differrent enums (upper vs. lower snake case etc.), you should use a uniform naming pattern there

dev/src/generate_sql_schema.py
```diff
@@ -1133,6 +1118,21 @@ def _first_to_second(t1: str, t2: str) -> bool:
...
+        to: str | None, reference: str | None
+    ) -> str:
+        if reference:
+            result = InternalHelper.ref_compiled.search(reference)
+            if result is None:
+                return reference.strip()
+            re_groups = result.groups()
+            return re_groups[0]
+        elif to:
+            return to.split("/")[0]
```
jsangmeister, line 1132: Use the `KEYSEPARATOR` constant here

dev/tests/test_generic_relations.py
```diff
@@ -37,6 +37,15 @@ def test_one_to_one_pre_populated_1r_1t(self) -> None:
             meeting_row = DbUtils.select_id_wrapper(curs, "meeting", self.meeting1_id, ["default_meeting_for_committee_id"])
             assert meeting_row["default_meeting_for_committee_id"] == self.committee1_id
 
+    # TODO: remove test, fiktiv test with 1r:1tR, test s.o, in der die meeting-Seite ein SQL hat
+    # jetzt setze ich mal ein erequired auf der Meeting Seite. Was ist übrigens mit 1r:1rR
```
jsangmeister, line 41: remove

jsangmeister (changes requested):

r-peschke:
> Small typos or clarifications. The concept itself looks good. I'm missing a list of required attributes for reach relation type, e.g.:
> 
>     * 1:1 must always contain a `reference` on one side and a `sql` on the other
> 
>     * similar rules for 1:n and m:n
> 
> 
> (just an example because I don't know the actual rules yet)
> 
> This can then be validated if we want via the `to` attributes, which is another reason why I would like to keep them.

This is documented in models.yml as list of implemented from-to-rules.

dev/src/generate_sql_schema.py
r-peschke: done

dev/src/generate_sql_schema.py
r-peschke: Used the CapWords or CamelCase style, recommended by PEP8 for type variables and class names.

r-peschke:
> Small typos or clarifications. The concept itself looks good. I'm missing a list of required attributes for reach relation type, e.g.:
> 
>     * 1:1 must always contain a `reference` on one side and a `sql` on the other
> 
>     * similar rules for 1:n and m:n
> 
> 
> (just an example because I don't know the actual rules yet)
> 
> This can then be validated if we want via the `to` attributes, which is another reason why I would like to keep them.

done in models.yml

models.yml
r-peschke: done

models.yml
r-peschke: done, including next word

models.yml
r-peschke: done

.gitignore
r-peschke: done

models.yml
r-peschke: done

jsangmeister (approved):

PR #25, opened 2024-02-09
Create PRs in all depending repos

### LINKED ISSUE #23: Open pull requests in all related repositories after an update
After this repo has been updated, pull requests should be opened automatically in all relevant repos (currently backend, client, autoupdate and search service) which update the respective submodule. This way, it is ensured that
1) the maintainers of the repo are informed about the changes
2) all repos always share the same version of the meta repo (assuming that most of the time the PR can simply be merged directly), which can be checked in the main repo
3) the overhead of updating the modules and creation of PRs is kept as low as possible.

Sometimes, manual intervention will of course be necessary, but this should be the regular case.

jsangmeister:
resolves #23


jsangmeister:
Changed the structure so the this repository handles the PR creation and file updates. This reduces the overhead in other repos, although one could argue that the meta repo should not have to care about how the autoupdate service generates its files, but I think it's okay to do it this way.

peb-adr (approved):
A tad bit skeptical about some hardcoded values and usernames, but making it dynamic is probably not worth the effort.

Issue #2, opened 2024-02-05
Add validator for the `search.yml`


jsangmeister:
The validator should check that all given collections and fields are actually valid.

PR #39, opened 2024-02-28
Fix searches including users with stucture levels failing

bastianjoel:
If a search result is containing a user with a structure level the search is currently failing because the relation type is wrong.


jsangmeister:
An implementation of https://github.com/OpenSlides/openslides-meta/issues/2 would have caught this ;)

jsangmeister (approved):

PR #53, opened 2024-03-12
Adapt validate.py for new fields

### LINKED ISSUE #49: Adaption validate.py to relational database design
The validation-script should pass the tests with the current models.yml. There are some new attributes like reference, deferred and unique. I'm sorry, but there will be more to check.
There is a make target defined as validate-models.
There is a debugger activated, to work you should copy the following code into youtrlaunch.json (VS-Code, Codium):
        {
            "name": "Python: attach META 5678",
            "type": "debugpy",
            "request": "attach",
            "connect": {
                "host": "localhost",
                "port": 5678
            },
            "justMyCode": false,
            "pathMappings": [
                {
                    "localRoot": "${workspaceFolder}/openslides-backend/global/meta",
                    "remoteRoot": "/app"
                },
            ]
        },

luisa-beerboom:
Closes #49

Couldn't find documentation about what rules relate to the new fields, so I laboured under the assumption that everything in the models.yml was formatted correctly. I doubt that this is actually the case since I also found some type errors in it which I _haven't_ fixed.

Based on what I saw in the models.yml I made the following assumptions about the new rules, which I am unsure about:
- Relation fields need to have at least one of `to`, `reference` and `sql`
- Other number fields may also have `to` or `sql`

I have no idea how the `unique` attribute that was hinted at is supposed to work as it is no where in the models.yml, so I didn't build in anything related to that for now

@r-peschke can you give some more information?


r-peschke:
Talked about it and decided to close in favor of https://github.com/OpenSlides/openslides-meta/pull/84

PR #54, opened 2024-03-13
Implement new los setting for secondary contributions

luisa-beerboom:
See OpenSlides/openslides-backend#2280

### DIFF
--- a/models.yml
+++ b/models.yml
@@ -1011,6 +1011,10 @@ meeting:
     type: boolean
     default: False
     restriction_mode: B
+  list_of_speakers_hide_secondary_contributions_count:
+    type: boolean
+    default: False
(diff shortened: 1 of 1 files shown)


models.yml
```diff
@@ -1011,6 +1011,10 @@ meeting:
     type: boolean
     default: False
     restriction_mode: B
+  list_of_speakers_hide_secondary_contributions_count:
```
jsangmeister, line 1014: Mh, I think something like `hide_contribution_numbering` or `hide_contribution_count` would fit better - I don't think "secondary" is a good descriptor for all contributions after the first...

jsangmeister (changes requested):

jsangmeister (approved):

PR #58, opened 2024-03-15
Change restriction mode of `user/last_email_sent`

jsangmeister:


jsangmeister:
see https://github.com/OpenSlides/OpenSlides/issues/6723

PR #59, opened 2024-03-15
Change restriction modes of username and saml_id

jsangmeister:


jsangmeister:
see https://github.com/OpenSlides/OpenSlides/issues/6723

PR #65, opened 2024-03-25
Readd 'to' relation fields

jsangmeister:
As discussed, this adds back the removed `to` fields for relations. With these changes, both the `generate-models` as well as the `generate-relational-schema` work without errors.


r-peschke:
> As discussed, this adds back the removed `to` fields for relations. With these changes, both the `generate-models` as well as the `generate-relational-schema` work without errors.

They work without errors, but both of the results are faulty or/and incomplete. It is just a way we COULD use the programmatic relation handling

jsangmeister:
> They work without errors, but both of the results are faulty or/and incomplete. It is just a way we COULD use the programmatic relation handling

Where are those errors? Regarding the incomplete information, I excluded the new relational attributes because I don't know yet which of these we'll actually need. I plan to implement them as the writer adapter gets updated and it becomes clear what information is required.

r-peschke (commented):
We should merge or use this in #17, because the results are not correct. The resulting models.py will be nearly or perhaps 100% identical to the old one and this is wrong.

jsangmeister:
> We should merge or use this in #17, because the results are not correct.

I think we should rather merge #17. We should use the feature branch, if possible, and merge changes as soon as they are required or usable by another actor. It is okay to merge unfinished or not correctly working code into the feature branch as it is still undergoing heavy development and the current status of the feature branch is far from final anyway.

> The resulting models.py will be nearly or perhaps 100% identical to the old one and this is wrong.

Why is this wrong? For now, it suffices. When the writer adapter gets changed, we will need some additional information in the `models.py` and can then add exactly what we need.

r-peschke:
Closing this, because merged the models.yml into #17

PR #73, opened 2024-04-09
use sql only in relaton fields

r-peschke:
- attribute sql only used for relation fields
- error/warning reporting on stdout and inside schema file
- fixed table names (now with "_t"-appendix)

### DIFF
--- a/dev/src/generate_sql_schema.py
+++ b/dev/src/generate_sql_schema.py
@@ -26,11 +26,13 @@ class SchemaZoneTexts(TypedDict, total=False):
 
     table: str
     view: str
+    post_view: str
     alter_table: str
     alter_table_final: str
     create_trigger: str
     undecided: str
     final_info: str
+    errors: list[str]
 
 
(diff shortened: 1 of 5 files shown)


dev/src/generate_sql_schema.py
```diff
@@ -76,7 +78,9 @@ class GenerateCodeBlocks:
     )  # Key=Name, data: collected content of table
 
     @classmethod
-    def generate_the_code(cls) -> tuple[str, str, str, str, str, list[str], str, str]:
+    def generate_the_code(
+        cls,
+    ) -> tuple[str, str, str, str, str, list[str], str, str, list[str]]:
```
jsangmeister, line 83: Not specific to this PR, but this should really be a custom class/typed dict. Such a convoluted tuple is not understandable in any way

jsangmeister (commented):
I don't really see how to review this appropriately  as there are many fine-granular changes. If you think this is fine, feel free to merge it

PR #83, opened 2024-04-17
use python-sql instead of psycopg.sql for generating sql-queries

r-peschke:


dev/src/db_utils.py
```diff
@@ -1,83 +1,38 @@
-from typing import Any, cast
+from typing import Any
 
-from psycopg import Cursor, sql
+from sql import Column, Table  # type: ignore
```
jsangmeister, line 3: Add the following section to the `setup.cfg` to not have to add `# type: ignore` everytime you import something from the `sql` package:
```
[mypy-sql]
ignore_missing_imports = true
```

jsangmeister (commented):
It would've been nice if you'd have separated the commits into one single "cleanup" commit and one commit with actual changes. This way, it is really hard to review.

Apart from that, I don't see many changes which are not simple reformattings. I noticed that there are still many usages of `psycopg.sql` in `dev/tests/base.py`, is this intended or do you plan to clean that up as well?

r-peschke:
> It would've been nice if you'd have separated the commits into one single "cleanup" commit and one commit with actual changes. This way, it is really hard to review.
> 
> Apart from that, I don't see many changes which are not simple reformattings. I noticed that there are still many usages of `psycopg.sql` in `dev/tests/base.py`, is this intended or do you plan to clean that up as well?

It is intended to use some psycopg.sql, because python-sql has some limitations:

- only dml-, but no ddl-support
- only implementation for standard sql. Some special features of postgres are not implemented intentionally

IMO python-sql is helpful to write the usual dml-sql, for special cases we need the (harder) psycopg.sql

dev/src/db_utils.py
r-peschke: That's what I did yesterday and now tried again, but the errors stayed.

dev/setup.cfg
```diff
@@ -17,3 +17,12 @@ extend-ignore = E203,E501
...
 disallow_untyped_defs = true
+
+[mypy-sql.*]
+ignore_missing_imports = true
+
+[mypy-python_sql.*]
+ignore_missing_imports = true
+
+[mypy-helper_get_names.*]
+ignore_missing_imports = true
```
jsangmeister, line 28: These can be removed now, can't they?

dev/src/db_utils.py
```diff
@@ -1,83 +1,38 @@
-from typing import Any, cast
+from typing import Any
 
-from psycopg import Cursor, sql
+from src.python_sql import Column, Table
```
jsangmeister, line 3: Better would be `from .python_sql import ...`

dev/src/python_sql.py
```diff
@@ -0,0 +1,2 @@
+# imports from sql for usage with type: ignore, because mypy don'T recognize the original import
```
jsangmeister, line 1:
```suggestion
# mypy does not recognize the imports from `python-sql` correctly. Therefore, we gather them in this file so they can be imported from here in other places without using `type: ignore` everytime.
```

jsangmeister (changes requested):

dev/setup.cfg
r-peschke: done

dev/src/db_utils.py
r-peschke: done

jsangmeister (approved):

PR #104, opened 2024-05-29
New chyron color fields in projector

luisa-beerboom:
See OpenSlides/openslides-backend#2437

### DIFF
--- a/models.yml
+++ b/models.yml
@@ -3874,7 +3874,15 @@ projector:
     type: color
     default: "#317796"
     restriction_mode: A
+  chyron_background_color_2:
+    type: color
+    default: "#134768"
+    restriction_mode: A
   chyron_font_color:
+    type: color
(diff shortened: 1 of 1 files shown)


models.yml
```diff
@@ -3874,10 +3874,18 @@ projector:
     type: color
     default: "#317796"
     restriction_mode: A
+  chyron_background_color_2:
+    type: color
+    default: "#134768"
```
reiterl, line 3879: Where does this default come from?

reiterl (commented):
I am not sure, if the default is correct.

models.yml
luisa-beerboom: Same colour as the global headbar in default theme

reiterl: Okay, I think it should be fine then. Nice.

reiterl (approved):
LGTM

Issue #125, opened 2024-06-28
Meeting.create: make duplicate from (optionally) mandatory


Elblinator:
**Wanted new field:**
New Orga wide setting to make the ‚duplicate from‘ field obligatory
The new setting should be a boolean, Orga and Superadmins can activate this setting

PR #129, opened 2024-07-12
Add orga setting for duplicate-from mandatory

reiterl:
closes https://github.com/OpenSlides/openslides-meta/issues/125

If this setting is set, committee admins must choose a template as duplicated-from at meeting create.

### DIFF
--- a/models.yml
+++ b/models.yml
@@ -142,6 +142,9 @@ organization:
     restriction_mode: A
     enum: *languages
     required: true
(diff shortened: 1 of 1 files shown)


luisa-beerboom (commented):
I think `require_duplicate_from` would be a better name, this one is a bit confusing

reiterl:
Updated.

PR #133, opened 2024-07-23
Make changes for locking out participants

### LINKED ISSUE #120: Add new attribute for locked_out feature
See https://github.com/OpenSlides/OpenSlides/issues/6756 for reference

The meeting_user needs a new Attribute. 
E.g. is_locked_out : boolean
A person needs the permission user.canManage to be able to lock out other participants
A person can see the attribute with the user.canSeeSensitiveData permission.

luisa-beerboom:
Closes #120

See Openslides/openslides-backend#2498

### DIFF
--- a/models.yml
+++ b/models.yml
@@ -408,7 +408,7 @@ meeting_user:
     restriction_mode: A
   locked_out:
     type: boolean
-    restriction_mode: A
+    restriction_mode: E
 
   user_id:
     type: relation
--- a/models.yml
+++ b/models.yml
@@ -406,6 +406,9 @@ meeting_user:
     type: decimal(6)
     minimum: "0.000001"
     restriction_mode: A
+  locked_out:
+    type: boolean
+    restriction_mode: A
 
   user_id:
     type: relation


models.yml
```diff
@@ -406,6 +406,9 @@ meeting_user:
     type: decimal(6)
     minimum: "0.000001"
     restriction_mode: A
+  locked_out:
+    type: boolean
+    restriction_mode: A
```
ostcar, line 411: Restriction mode A means, that it can be seen by many users.

Restriction mode D would be for admins.

I don't know what is better in this case

See https://github.com/OpenSlides/openslides-autoupdate-service/blob/93f1494a6b5c581f98f7401aefa2a9368b4eb091/internal/restrict/collection/meeting_user.go#L11-L22

ostcar (approved):

models.yml
luisa-beerboom: @Elblinator

Elblinator: Could it be done to create an additional permission mode which is doing the same as the permission mode B for user? As far as I can see the permission mode B (for user) is allowing the use to see the information of they have the permission canSeeSensitiveData

ostcar: I am not sure what you mean. user/b is a bit more complicated: https://github.com/OpenSlides/openslides-autoupdate-service/blob/93f1494a6b5c581f98f7401aefa2a9368b4eb091/internal/restrict/collection/user.go#L34-L39

And user/b is not meeting specific,

Do you mean, that we create a new restriction for that field, that only users can see it, that have the permission users.can_see_sensible_data? This easy and possible. I am against using that permission, but this is only my opinion and not a technical problem.

If you want to do this, than change the restriction_mode to a new letter.

(By the way, the restriction mode do not have to be a letter. It can be a string. It could be: "like_user_b" or "only_with_senisbile_data_permisson")

Elblinator: > Do you mean, that we create a new restriction for that field, that only users can see it, that have the permission users.can_see_sensible_data?

Yes, exactly

luisa-beerboom: Done, it's now restriction mode E

ostcar: I am sorry, I was in the wrong feature. I was thinking about locked meeting when I posted my comment. I though, we were talking about the meeting collection.

In the meeting_user collection, we already have the mode D:
```
// Mode D: Y can see these fields if at least one condition is true:
//
//	The request user has the OML can_manage_users or higher.
//	The request user has user.can_manage in the meeting
```
The field, that is using this mode is `meeting_user/comment`.

I would suggest, that we put the locked_out field in the restriction mode D. It is not the same as we discussed here, but I think, it is better to have less cases. Or there can be confusion in the client which kind of admin/user can see what.

Elblinator: Restriction Mode D is too restrictive
Restriction Mode D --> you need user.can_manage
hard criteria for the feature = user with user.can_update need to be able to change the locked_out status
soft criteria = user with user.can_see_sensitive_data need to see if a user is locked_out

If we talk and (sub-sequentially) change the behaviour for user.can_see_sensitive_data soon then the soft criteria **could** change.
My gut feeling is that we will need the restriction mode for user.can_see_sensitive_data even after that discussion. please implement it like that

ostcar: Ok. Its implemented in https://github.com/OpenSlides/openslides-autoupdate-service/pull/972

PR #146, opened 2024-08-09
Make changes for public organization mediafiles

### LINKED ISSUE #121: Allow orga-files/folders to be accessible in meetings
**Desired behaviour:**
- If a file/folder is public then this file/folder needs to be displayed in every meeting
- Files/folders need a new flag to show that they are public.

luisa-beerboom:
Closes #121

### DIFF
--- a/models.yml
+++ b/models.yml
@@ -3708,6 +3708,13 @@ mediafile:
   token:
     type: string
     restriction_mode: A
+  is_public:
+    type: boolean
+    restiction_mode: A
+  public_in_organization_id:
+    type: relation
+    to: organization/public_mediafile_ids
+    restriction_mode: A
   parent_id:
     type: relation
     to: mediafile/child_ids
@@ -3730,10 +3737,6 @@ mediafile:
     type: relation-list
     to: meeting_mediafile/mediafile_id
     on_delete: CASCADE
-    restriction_mode: A
-  public_in_organization_id:
-    type: relation
-    to: organization/public_mediafile_ids
     restriction_mode: A
 
 meeting_mediafile:
--- a/models.yml
+++ b/models.yml
@@ -1968,12 +1968,12 @@ group:
     to: meeting/anonymous_group_id
     on_delete: PROTECT
     restriction_mode: A
-  mediafile_access_group_ids:
+  meeting_mediafile_access_group_ids:
     type: relation-list
     to: meeting_mediafile/access_group_ids
     equal_fields: meeting_id
     restriction_mode: A
-  mediafile_inherited_access_group_ids:
+  meeting_mediafile_inherited_access_group_ids:
     type: relation-list
     to: meeting_mediafile/inherited_access_group_ids
     description: Calculated field.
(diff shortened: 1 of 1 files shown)


models.yml
```diff
@@ -2196,7 +2205,7 @@ list_of_speakers:
         - motion_block
         - assignment
         - topic
-        - mediafile
+        - meeting_mediafile
```
r-peschke, line 2208: Why you can't stay with the mediafile? The permission was checked on assigning the mediafile to an object.
What kind of permission will be checked on usage of the object? If user A has the right to see a mediafile, he can assign the mediafile to a list_of_speakers and to a motion.
If user B is not allowed to see this mediafile, is he forced to see the list_of_speakers without the grafik-icon? On a motion's attachment this may have a sense, on speakers-list not, assuming the mediafile is used as a kind of icon for the speakers list or another object.

models.yml
```diff
@@ -2393,7 +2402,7 @@ topic:
 
   attachment_ids:
     type: relation-list
-    to: mediafile/attachment_ids
+    to: meeting_mediafile/attachment_ids
```
r-peschke, line 2405: see above

models.yml
```diff
@@ -2673,7 +2682,7 @@ motion:
     restriction_mode: C
   attachment_ids:
     type: relation-list
-    to: mediafile/attachment_ids
+    to: meeting_mediafile/attachment_ids
```
r-peschke, line 2685: see above

models.yml
```diff
@@ -3574,7 +3583,7 @@ assignment:
     restriction_mode: A
   attachment_ids:
     type: relation-list
-    to: mediafile/attachment_ids
+    to: meeting_mediafile/attachment_ids
```
r-peschke, line 3586: see above

models.yml
```diff
@@ -3705,24 +3714,12 @@ mediafile:
...
-  is_public:
-    type: boolean
-    description: "Calculated field. inherited_access_group_ids == [] can have two causes: cancelling access groups (=> is_public := false) or no access groups at all (=> is_public := true)"
-    read_only: true
-    required: true
-    restriction_mode: A
   token:
     type: string
     restriction_mode: A
-  inherited_access_group_ids:
```
r-peschke, line 3717: That means on migration you have to migrate all the data to newly created meeting_mediafile instances

models.yml
```diff
@@ -3734,6 +3731,49 @@ mediafile:
...
+    required: true
+    restriction_mode: A
+  is_public:
+    type: boolean
+    description: "Calculated in actions. inherited_access_group_ids == [] can have two causes: cancelling access groups (=> is_public := false) or no access groups at all (=> is_public := true)"
+    required: true
+    restriction_mode: A
+  inherited_access_group_ids:
+    type: relation-list
+    to: group/mediafile_inherited_access_group_ids
```
r-peschke, line 3770: What do you think of explaining the calculation rules here?

r-peschke (commented):
Only some questions/remarks

models.yml
luisa-beerboom: Good idea, will do just that.

models.yml
luisa-beerboom: Considering I have already done that in the backend PR, what point are you trying to make with this comment?

models.yml
luisa-beerboom: I don't get what this has to do with permissions, however the reasons I moved all these fields to their own model are:
1. To make backend-side meeting-related-mediafile-data-management easier. If, for example, a published orga mediafile's access groups are edited in a meeting, the re-calculation of inherited access groups for this mediafile and its descendants doesn't call for reconsideration of the inherited access groups that are set in any other meetings. It would make no sense at all to juggle endlessly long lists of access groups even though usually only about 3 of them are relevant and may need to be changed at a time.
2. To make client-side display easier. For example: If the client wants to display in the mediafile list, whether a mediafile is being projected in this meeting, he shouldn't have to filter through a list of 300 current projections just to see if one of them is from this meeting.
3. To keep integrity checks easier. Using the meeting-internal `meeting_mediafile` model allows us to keep the `equal_fields: meeting_id` requirement for these relations, which would definitely have to be removed, if we used the base `mediafile` model since, due to the new "publishing" feature, it can now have an organization owner, even if it is used in a meeting context. However `meeting_mediafile` _always_ has a distinct `meeting_id`

Does that answer your questions?

models.yml
r-peschke: Just to not forget! This was the first I checked in reviewing the backend side, but then I forget to delete this here.

bastianjoel:
There is something wrong with the calculation of `inherited_access_groups`. 
I set access groups of a published file with a folder in a meeting. The `access_group_ids` were updated but `inherited_access_group_ids` were not updated. 

Editing the parent folder also did not change the `inherited_access_group_ids` of that file.

PR #167, opened 2024-09-17
limit of field name length documented in modesl.yml

r-peschke:


ostcar:
@r-peschke Could you explain, why the 25 characters are necessary? I guess, this is some restriction for the future SQL change. But why exactly 25 characters? How did you come up with this number?

r-peschke:
> @r-peschke Could you explain, why the 25 characters are necessary? I guess, this is some restriction for the future SQL change. But why exactly 25 characters? How did you come up with this number?

Up to now the problem of variables came up some times. Last before the current we had with variables including list_of_speaker as part of their names.
Currently we struggled with the new table meeting_mediafile. That leads to field names like `meeting_mediafile_inherited_access_group_ids`.
The limit of postgres are 63 characters for each type of name.
Now the part of the relational db:
We have lots of n:m relations, that will be resolved by an intermediate table in a generic way. The name convention for this table is `nm_{name of table 1}_{field name of table 1}_{name of table 2}`. Discounting the 5 constant chars 58 chars remain for the junction of the 3 names, 19.3 per name. Because a column name often includes the table name, especially in relation-fields, we expected them to be longer than the 2 table_names. So 25 chars seems to be a good compromise.

For not changing **now** a lot of field names on the main branch with all the consequences, we introduced an algorithm that uses the initial letters  of the `_`-connected words of the field name, if it has more than 25 chars, see #169. This reduces the readability of the db-design and in time we want to shorten the names.

PR #204, opened 2025-01-11
Add fields for crypto_vote

ostcar:


ostcar:
This does not represent the current concept

PR #221, opened 2025-02-24
No longer set milestone for 'update-meta' PRs

peb-adr:
This was a hardcoded id until now and was error prone since the beginning. It resulted in non-sensible milestones being set (due to inconsistent IDs accross repos).
Most recently it caused abortion of the actions as openslides-meta started using meta as submod but having no milestones defined.


peb-adr:
@rrenkert FYI, as discussed.

PR #229, opened 2025-03-06
WIP: back-rel trigger

peb-adr:
This is a WIP generating triggers for the simplest of the four relation types `relation`.
`relation-list`, `generic-relation`, `generic-relation-list` are yet to be done.

So this is in no way complete!


peb-adr:
Maybe consider not having an extra postgres function `notify_modified_related_models`.
Instead it might be nicer to have a single more compatible `notify_...` function which can be called for all notifying triggers.

LinKaKling:
i continued this in #256

PR #231, opened 2025-03-10
Add origin motion toggle meeting settings

### LINKED ISSUE #225: Add two new settings for motions view
Add new setting in motions > view: Activate origin view. / We need a setting to deactivate the new feature by default. Most meetings won't need it and therefore we should integrate an option to deactivate it by default.

Add new setting in motions > view: Add a new option to toggle default behaviour of the slide controller.

We should discuss whether these need to be settings for the organization or the meeting.

bastianjoel:
resolves #225

### DIFF
--- a/models.yml
+++ b/models.yml
@@ -1124,6 +1124,14 @@ meeting:
     type: boolean
     default: False
     restriction_mode: B
+  motions_origin_motion_toggle_default:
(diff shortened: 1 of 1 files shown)


hjanott (approved):
lgtm.

PR #240, opened 2025-04-02
Trim table suffix in notify trigger

rrenkert:


ostcar:
@peb-adr could we merge this for now, so I can use it?

PR #248, opened 2025-04-04
Fix decimal type

ostcar:
Fix the decimal type in postgres.

Also removed the decimal type without the size 6. We don't have any other size and we should definitely not introduce one. For the future, we should rename `decimal(6)` to `decimal`


rrenkert (approved):
We should think about implement the decimal precision more generic. For now this is ok since the documentation in the models.yml defines that we only support decimal(6).

PR #249, opened 2025-04-04
Send notifications on transaction commit

rrenkert:


dev/sql/schema_relational.sql
```diff
@@ -48,26 +59,61 @@ begin
...
+$log_notify_trigger$ LANGUAGE plpgsql;
+
+CREATE FUNCTION notify_modified_models() RETURNS trigger AS $notify_trigger$
+DECLARE
+    pl TEXT;
+    body_content_text TEXT;
+BEGIN
+    -- Running the trigger for the first time in a transaction the table is created and dropped after commiting the transaction.
+    -- Every next run of the trigger in this transaction raises a notice that teh table exists. This is not ideal.
+    CREATE LOCAL TEMPORARY UNLOGGED TABLE
```
ostcar, line 85: When applying the schema, I get an error in this like:

`ERROR: syntax error at or near "UNLOGGED" (SQLSTATE 42601)`

On the Postgres Docu, it seems, that `TEMPORARY` and `UNLOGGED` can not be used together. There is a `|` in the documentation between them.

I don't have experience with either of the attributes. But it seems, that a temporary table is also unlogged. So the `UNLOGGED` attribute can be removed:

https://stackoverflow.com/a/59247206

There is a typo in the docstring above this line `teh` => `the`

dev/sql/schema_relational.sql
```diff
@@ -48,26 +59,61 @@ begin
...
+                    WHERE xact_id = pg_current_xact_id()),
+                '","')
+            INTO body_content_text;
+        -- Concat current transaction id and formated fqid list to a json object.
+        pl := '{"xactId":' ||
+            pg_current_xact_id() ||
+            ',"fqids":["' ||
+            body_content_text ||
+            '"]}';
+        PERFORM pg_notify('os-notify', pl);
```
ostcar, line 107: I am not able to listen on 'os-notify'. It seems that `-` is not an allowed character. Could you rename this to `os_notify` or some other name?

The documentation says, the channel has to be `any identifier`. And "the internet" says, and postgres identifier can only be letters, numbers and underscore.

ostcar (changes requested):
I am not finish testing it. But it seems to work. All fqids of a transaction are send to the message bus at once.

It still has the bug, that the table name (with `_t` suffix) is used.

ostcar:
Could you also merge this on the current HEAD of the branch? It contains the decimal fix. I have tests that fail, when the decimal fix in not included.

ostcar:
I can confirm, that this works. With the necessary fixes mentioned above, the autoupdate applies all changes in one transaction in one run.

dev/sql/schema_relational.sql
rrenkert: Done. Added the `UNLOGGED` shortly before going into weekend. ;)

dev/sql/schema_relational.sql
rrenkert: Done. postgres itself seems to support `-`. Used psql to test this PR and had no problems with it.

PR #251, opened 2025-04-06
Add/Fix file with init data

ostcar:


rrenkert (changes requested):
Please use a better name for the `init_data.sql` file like `base_test_data.sql` or `init_test_data.sql` beacuse it contains test data that is created by users in regular instances.

Did not test the new workflow action but looks good to me.

ostcar:
As discussed with @rrenkert, it can be merged

PR #252, opened 2025-04-08
remove _t from table name in notification to get correct fqid

LinKaKling:
The notification contains the correct fqids now.
eg. "user/1" instead of "user_t/1"


dev/src/generate_sql_schema.py
```diff
@@ -809,8 +811,9 @@ def get_view_body_end(table_name: str, code: str) -> str:
     @staticmethod
     def get_notify_trigger(table_name: str) -> str:
         own_table = HelperGetNames.get_table_name(table_name)
+        escaped_table_name = '\'' + table_name + '\''
```
hjanott, line 814: This line leads to an error with black.
I suggest using an f-string or double quotes.

hjanott (changes requested):

ostcar:
I can confirm, that it works. The fqids in the notify message are correct

rrenkert (approved):

PR #253, opened 2025-04-09
This is an example, how a splitted models.yml could look like

ostcar:


ostcar:
Done in https://github.com/OpenSlides/openslides-meta/pull/365

PR #254, opened 2025-04-09
Log modified fqids and send transaction id on transaction end

rrenkert:


LinKaKling (approved):

ostcar (approved):
It works. I merge this, because I already wrote the go code to test this

PR #255, opened 2025-04-10
Generate views for all collections

hjanott:
Simple collections didn't get their views yet and that would lead to problems for the database reader when reading on those collections.


LinKaKling (commented):
Code looks fine.

Issue #228, opened 2025-03-06
[rel-db] Add notify trigger for back relations


peb-adr:
https://github.com/OpenSlides/openslides-meta/pull/212

is merged now, such that autoupdate development can start using the basic triggers.
However more triggers covering changes for the back-relation fields are still to be generated.

For the different relation types the schema generator already has conditional handling.
In particular the methods

    get_relation_type
    get_relation_list_type
    get_generic_relation_type
    get_generic_relation_list_type

are of interest.

Within them information for appropriate trigger definitions is available.

peb-adr:
I did have some initial progress on this matter, see https://github.com/OpenSlides/openslides-meta/pull/229

LinKaKling:
I had some further progress in https://github.com/OpenSlides/openslides-meta/pull/256

PR #256, opened 2025-04-11
Related models notification log trigger

LinKaKling:
Adds a Trigger that logs changes of foreign key/relation columns  as changes of the relation_list in the corresponding table.
+quick fixed base_data.sql
its a continuation of peb-adr:pg-notify-v2 for https://github.com/OpenSlides/openslides-meta/issues/228


dev/src/generate_sql_schema.py
```diff
@@ -693,6 +700,30 @@ class Helper:
...
+            operation:= LOWER(TG_OP);
+            ref_column := TG_ARGV[1];
+            foreign_table := TG_ARGV[0];
+
+            EXECUTE format('SELECT $1.%s', ref_column) INTO foreign_id USING NEW;
+            IF (TG_OP = 'DELETE') THEN
+                EXECUTE format('SELECT $1.%s', ref_column) INTO foreign_id USING OLD;
+            END IF;
+
+            fqid := foreign_table || '/' || foreign_id;
```
rrenkert, line 720: The `foreign_id` is NULL if a collection is inserted or updated and a not required field is empty, e.g. insert a user with username only (gender_id is empty). The trigger is executed and inserts a row into os_notify_log_t with an empty fqid.

rrenkert (changes requested):

dev/src/generate_sql_schema.py
```diff
@@ -883,6 +915,34 @@ def get_on_action_mode(action: str, delete: bool) -> str:
...
+        )
+        own_table = HelperGetNames.get_table_name(table_name)
+        return FOREIGN_KEY_NOTIFY_TRIGGER_TEMPLATE.substitute(
+            {
+                "trigger_name": trigger_name,
+                "own_table": own_table,
+                "ref_column": ref_column,
+                "foreign_table": foreign_table,
+            }
+        )
```
hjanott, line 944: I think this is more readable and shorter since you can delete lines 928 through 932 also and the concatenation will happen in one place.
```suggestion
        return f"CREATE TRIGGER {trigger_name} AFTER INSERT OR UPDATE OF {ref_column} OR DELETE ON {own_table}\n FOR EACH ROW EXECUTE FUNCTION log_modified_related_models('{foreign_table}', '{ref_column}');\n"
```

hjanott (commented):
One minor remark.

dev/src/generate_sql_schema.py
ostcar: I also run into this problem. I did not see this comment and it took me some time to debug it.

Please change the os_notify_log_t table (line 105) to:
```sql
CREATE TABLE os_notify_log_t (
    id integer PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
    operation varchar(32),
    fqid varchar(256) NOT NULL,
    xact_id xid8,
    timestamp timestamptz
);
``` 

(add the `NOT NULL` constain to the fqid column).
This prefents postgres to add invalid data to the os_notify_log_t table

ostcar (changes requested):
I can currently not test this, if it create null values on fqid in the notify table

LinKaKling:
ToDo:
  -add Notification trigger for nm and gm tables
  -prevent multiple logging of one fqid per transaction? 
  -cleanup

dev/src/generate_sql_schema.py
```diff
@@ -120,6 +120,8 @@ def generate_the_code(
         for table_name, fields in MODELS.items():
             if table_name in ["_migration_index", "_meta"]:
```
hjanott, line 121: I'm thinking if we should call this variable collection.

dev/src/generate_sql_schema.py
```diff
@@ -325,6 +327,14 @@ def get_relation_type(
                     initially_deferred,
                 )
             )
+            table_name = HelperGetNames.get_table_name(table_name)
```
hjanott, line 330: Should this be called collection_name, also?
Just a quick thought to consider. Maybe we can remove this function completely by creating the `TableFieldType` for own_table_field and foreign_table_field completely?

rrenkert (approved):

PR #257, opened 2025-04-15
Add motion_state setting for amendment forwarding

luisa-beerboom:


models.yml
```diff
@@ -3043,6 +3043,9 @@ motion_state:
     type: boolean
     restriction_mode: A
     default: false
+  allow_amendment_forwarding:
```
hjanott, line 3046: This does not need a default?

hjanott (approved):

models.yml
hjanott: It seems not. This way we don't need a migration.

Issue #258, opened 2025-05-07
[rel-db] Fix generate_sql_schema.py for generic views.


hjanott:
Generic fields in views currently deliver the id of the intermediate table rows instead of the fqids. I ran into this with the organization tags. The resulting ddl should look like this.

```
CREATE VIEW "organization_tag" AS SELECT *,
(select array_agg(g.tagged_id) from gm_organization_tag_tagged_ids g where g.organization_tag_id = o.id) as tagged_ids
FROM organization_tag_t o;
```

PR #263, opened 2025-05-20
Initially defer and use correct field in nm gm

### LINKED ISSUE #260: [rel-db] intermediate tables aren't initially deferred
Adding a row to an intermediate table like `nm_committee_user_ids_user` will immediately result in an error if one of the relations is not yet existing. This needs to be fixed in the code generation for the relational schema by adding `INITIALLY DEFERRED` like it has been done for other columns.

hjanott:
Closes #260, #258
* initially defer nm relations
* use correct field for generic nm relations
* introduce intermediate column for TableFieldType

### DIFF
--- a/dev/sql/schema_relational.sql
+++ b/dev/sql/schema_relational.sql
@@ -1168,20 +1168,20 @@ CREATE TABLE import_preview_t (
 -- Intermediate table definitions
 
 CREATE TABLE nm_meeting_user_supported_motion_ids_motion (
-    meeting_user_id integer NOT NULL REFERENCES meeting_user_t (id),
-    motion_id integer NOT NULL REFERENCES motion_t (id),
+    meeting_user_id integer NOT NULL REFERENCES meeting_user_t (id) INITIALLY DEFERRED,
+    motion_id integer NOT NULL REFERENCES motion_t (id) INITIALLY DEFERRED,
     PRIMARY KEY (meeting_user_id, motion_id)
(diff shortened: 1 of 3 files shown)


LinKaKling:
After this all REFERENCES are INITIALLY DEFERRED expect some in the gm tables.
Maybe we should change the GM_FOREIGN_TABLE_LINE_TEMPLATE for these tables in line 737 `"    ${gm_content_field} integer GENERATED ALWAYS AS (CASE WHEN split_part(${own_table_column}, '/', 1) = '${foreign_view_name}' THEN cast(split_part(${own_table_column}, '/', 2) AS INTEGER) ELSE null END) STORED REFERENCES ${foreign_table_name}(id) INITIALLY DEFERRED,"
(and generate the sql)

LinKaKling (changes requested):
fter this all REFERENCES are INITIALLY DEFERRED except some in the gm tables.
Maybe we should change the GM_FOREIGN_TABLE_LINE_TEMPLATE for these tables in line 737
`" ${gm_content_field} integer GENERATED ALWAYS AS (CASE WHEN split_part(${own_table_column}, '/', 1) = '${foreign_view_name}' THEN cast(split_part(${own_table_column}, '/', 2) AS INTEGER) ELSE null END) STORED REFERENCES ${foreign_table_name}(id) INITIALLY DEFERRED,"`
(and generate the sql)

PR #278, opened 2025-07-02
Add changes for history collection

luisa-beerboom:
See OpenSlides/openslides-backend#3009

### DIFF
--- a/models.yml
+++ b/models.yml
@@ -1957,6 +1957,7 @@ group:
         - agenda_item.can_see
         - agenda_item.can_see_internal
         - assignment.can_manage
+        - assignment.can_manage_polls
         - assignment.can_nominate_other
         - assignment.can_nominate_self
         - assignment.can_see
--- a/permission.yml
+++ b/permission.yml
@@ -14,6 +14,8 @@ assignment:
     can_manage:
         can_nominate_other:
             can_see:
+    can_manage_polls:
+        can_see:
     can_nominate_self:
         can_see:
 chat:
@@ -38,8 +40,6 @@ motion:
     can_manage:
         can_manage_metadata:
             can_see:
-        can_manage_polls:
-            can_see:
         can_see_internal:
             can_see:
         can_create:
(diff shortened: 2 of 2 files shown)


ostcar:
autoupdate: https://github.com/OpenSlides/openslides-autoupdate-service/pull/1229
openslides-go: https://github.com/OpenSlides/openslides-go/pull/65

luisa-beerboom:
> I made the relation from history_entry to its model required. I hope this was correct. If not, please revert

Reverted, because it is not supposed to be required. The entry is supposed to remain, potentially forever, even if the model is deleted, so the field needs to allow an empty value.

ostcar:
What about permissions in this case? If a motion exists, the object can be seen with the permission meeting.can_see_history. but if the motion gets deleted, the information is lost, to which meeting the element belongs. So a delete event can only be seen by a orga manager. 

If this is a problem, then we could add an optional meeting id to the history entry. This would be set for every entry on elements, that belong to a meeting, like a motion but would be none on organisation collections like a user. 

What do you think?

hjanott:
Since meeting_user changes are written into users history, the meeting_id could be useful there too.

luisa-beerboom:
The lazy solution would probably be to limit history of deleted models to orga admins.

As far as the suggestion goes, these are my thoughts:
- Users aren't meeting models but do have history entries how are they handled?
   - Especially since concerning users it is possible for there to be operations done concerning multiple meetings at the same time (think "add to multiple meetings" type stuff), all that information is going to end up in the same entry
 - The meeting_id field would still be empty if the meeting is deleted, for all models that were deleted so far (since we can't read out the information in the migration for those) and at the very least all pre-migration user entries. I guess this part can't be helped though
 - Do we want to allow committee admins to access their deleted meetings history? In that case, we'd need to hold that info as well

@rrenkert what do you think?

ostcar:
The restricter is updated for the new meeting_id field

luisa-beerboom:
@ostcar I just noticed that I added the restriction modes rather arbitrarily, perhaps you should take a look at that and maybe fix it

hjanott (approved):

PR #280, opened 2025-07-11
Move vote/user_token to restriction mode A

ostcar:
Its the last field in restriction mode B. After this, the restrictor for vote/B can be deleted.


ostcar:
This will be irrelevant with rel db

PR #288, opened 2025-07-17
[Rel-DB] Fix and unify the sql attributes

ostcar:
YAML has many different ways to represent multi line strings. The current version uses different onces. This PR unifies them.

[Here](https://stackoverflow.com/questions/3790454/how-do-i-break-a-string-in-yaml-over-multiple-lines) is a description, what is possible with yaml.

The current version also has a yaml-syntax-error, so it could not be pared by `openslides-go`

### DIFF
--- a/models.yml
+++ b/models.yml
@@ -399,7 +399,7 @@ user:
     restriction_mode: E
     read_only: true
     description: "Calculated field: Returns committee_ids, where the user is manager or member in a meeting"
-    sql:
+    sql: |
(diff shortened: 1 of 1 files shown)


hjanott (changes requested):
I like this idea. Please also remove those extra lines that make the sql look bloated (For instance after each comment.) from the `models.yml` and run `make generate-relational-schema` from the `dev` folder.

PR #292, opened 2025-07-28
Fix sql comments with quotes

hjanott:
Fixes what is mentioned here. https://github.com/OpenSlides/openslides-meta/pull/287#issuecomment-3121409649


hjanott:
I have changed the action scripts and as you can see I had to repair the base data so the tests go green.

PR #293, opened 2025-07-30
[rel-db] Handle relational fields referencing themselves

vkrasnovyd:
Example: motion.identical_motion_ids

**Behaviour before:** intermediate table creates relations created from both sides (1-2 and 2-1). Retrieving identical_motion_ids for motion/1 would result in [2, 2].

**New behaviour: ** only one pair (where id1 < id2) is processed, the other one is skipped. motion.identical_motion_ids returns correct ids list without the duplicates.


hjanott (approved):
Awesome job. I just have one little suggestion.

hjanott (approved):

PR #305, opened 2025-08-25
[Rel-DB] Remove option from base_data

ostcar:
The base data should only create necessary data to get an empty instance.

This PR removes the creation of an option. I don' think the option is necessary for a base instance.

The option is currently interfering with my changes to the vote-service. If it is not needed, then it would be best to remove it.


hjanott:
I created the test_data file and linted it with psqfluff. If you agree, I would also lint the other files the same way and add the linter to make files lint target as well as the schema generation.

dev/sql/test_data.sql
```diff
@@ -0,0 +1,190 @@
...
+    pollmethod,
+    onehundred_percent_base,
+    sequential_number,
+    content_object_id,
+    meeting_id
+)
+--rl:gr topic.poll_ids:poll.content_object_id
+VALUES (1, 'Titel21', 'analog', 'fast', 'YNA', 'disabled', 1, 'topic/1', 1);
+
+INSERT INTO nm_committee_manager_ids_user_t
```
LinKaKling, line 57: `INSERT INTO nm_committee_manager_ids_user_t (committee_id, user_id)`

LinKaKling (commented):
Do you want to delete base_data.sql?
I am not sure what base_data.sql should be. 
Something in between initial data and test_data? 
Anyways we can merge this and change or delete base_data in a new issue because removing the insert of option is more urgent i guess.

ostcar:
What I need is a base_data.sql that creates data needed in all tests, but nothing more.

For example, if I write the vote-service, I have to create a lot of tests that create polls on motions, topics, or assignments. To have a reasonable test, I want a database that doesn't have any motions, assignments, etc. in it, so I can manually create the test data that I need for the test case.

Here is an example of how such a test could look: https://github.com/OpenSlides/openslides-vote-service/blob/b3b420a89bc8b38b92262c4fdff37ae23c89ab34/vote/vote_test.go#L50-L69

But I don't want to create an organization, organization_tag, meeting for every test. That's why I want to have the base_data. It should create a meeting with id 1 and all the other collections that are necessary for that (like the organization).

If I understood [hjanott](https://github.com/hjanott) correctly, then you need a test file to test the SQL schema. For example, if n-m-tables work. That was the reason you added the option to the base_data.

I think these are two valid use cases. But they are different and therefore should not be in the same file. That is the reason why I think we need two files.

.github/workflows/test-sql.yml
```diff
@@ -40,7 +40,7 @@ jobs:
 
       - name: insert init data
         shell: bash
-        run: bash dev/scripts/apply_base_data.sh
```
ostcar, line 43: Could you also create a github-action, that tests the base_data.sql?

hjanott (approved):

PR #306, opened 2025-08-27
[rel-db] Introduce more relations/constraints / change handling of MODELS

hjanott:


vkrasnovyd:
Looks good and works for both 1-to-1 and lists.

I'd only suggest removing ` from relationship before %/%` part from the insert triggers. Since we pass empty string as `foreign_key` and `OLD.id` does not exist yet, we get message like `ERROR:  Trigger tr_i_meeting_default_projector_agenda_item_list_ids Exception: NOT NULL CONSTRAINT VIOLATED for meeting/2/default_projector_agenda_item_list_ids from relationship before <NULL>/`.

vkrasnovyd:
> I'd only suggest removing ` from relationship before %/%` part from the insert triggers.

As discussed, I'll take care of it in #308. The rest of the PR looks good to me.

PR #316, opened 2025-09-12
Permission poll progress

ostcar:
This is a proposal for a refactor of the poll permissions.

For a better diff, see: https://github.com/OpenSlides/openslides-meta/commit/b1caa741f6f3f7c9a2d5991633b59fa23b323fce

The old system used the permission poll.can_manage only for topic-polls. This proposal renames this to `agenda_item.can_manage_polls` what is more clear.

The old system used poll.can_see_progress for topic, motion and assignment polls. This means, that it was not possible to give a user only the permission for one of the collections. If we make a distinction for `can_manage_pools` we should also make a distinction for `can_see_progress`.

The old system has to problem, that if a user has the permission to create topic polls (poll.can_manage), he also gets the permission to see the progress bar for motion or assignment polls. This was probably not intended.

If you want, we can keep the old system for now, and I only change the permissions on the new-vote-service-branch

### DIFF
--- a/permission.yml
+++ b/permission.yml
@@ -10,12 +10,17 @@ agenda_item:
   can_manage:
     can_see_internal:
       can_see:
+  can_manage_polls:
+    can_see_progress:
+      can_see_internal:
+        can_see:
 assignment:
   can_manage:
     can_nominate_other:
       can_see:
   can_manage_polls:
-    can_see:
+    can_see_progress:
+      can_see:
   can_nominate_self:
     can_see:
 chat:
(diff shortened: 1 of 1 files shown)


ostcar:
With this change, this hidden feature is not necessary anymore: https://github.com/OpenSlides/openslides-autoupdate-service/blob/20b46419f3c193136b79261a6209c021bf804217/internal/restrict/collection/poll.go#L36

ostcar:
@bastianjoel @rrenkert : @MSoeb and I discussed this PR. We want to merge this in the new vote-service branch. Then, there will only be changes in the autoupdate-service, vote-service and the client. Is this ok?

bastianjoel:
@ostcar I am okay with doing this on the vote branch. However - I would prefer to wait with merging a bit longer until I got rid of all old unused poll code there. Otherwise this change might be a bit disruptive.

ostcar:
branch merged into https://github.com/OpenSlides/openslides-meta/pull/506

PR #318, opened 2025-09-16
Make changes for agenda forwarding

luisa-beerboom:
See OpenSlides/openslides-backend#3075

### DIFF
--- a/models.yml
+++ b/models.yml
@@ -765,6 +765,14 @@ committee:
     type: relation-list
     to: committee/forward_to_committee_ids
     restriction_mode: B
+  forward_agenda_to_committee_ids:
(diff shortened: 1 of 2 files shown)


models.yml
```diff
@@ -765,6 +765,14 @@ committee:
     type: relation-list
     to: committee/forward_to_committee_ids
     restriction_mode: B
+  forward_agenda_to_committee_ids:
```
hjanott, line 768: If this is an allow list, maybe it would be possible to represent this in the name. Adopting the schema of the back relation would result to 
```suggestion
  send_agenda_forwardings_to_committee_ids:
```
but describe more of a fixed recipient list.

PR #319, opened 2025-09-16
[rel-db] add sequential numbers before insert

### LINKED ISSUE #219: [rel-db] Use postgres sequence for `sequential_number` fields
Several collections have `sequential_number` fields.
These are currently incremented and set in the backend code.

It would be nice to alter the SQL schema such that these values are instead generated/incremented on DB level, most likely using sequences.

We have to ensure that these sequences are initialized correctly (i.e. `max(<collection>.sequential_number)`). This would be part of the corresponding migration in the backend.

hjanott:
Closes #219

### DIFF
--- a/dev/sql/schema_relational.sql
+++ b/dev/sql/schema_relational.sql
@@ -1,7 +1,7 @@
 
 -- schema_relational.sql for initial database setup OpenSlides
 -- Code generated. DO NOT EDIT.
--- MODELS_YML_CHECKSUM = '65aafbd8efddad3a70265441cbf116b1'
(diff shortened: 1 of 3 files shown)


ostcar (approved):
I cannot say anything about SQL. I will try it on the new vote service branch, when you merged it.

vkrasnovyd (approved):
Works and looks good.

vkrasnovyd:
@hjanott looks like there's no related PR with updated `models.py` in the backend.

vkrasnovyd:
During the backend cleanup found manual setting of `sequential_number` in the `meta/dev/tests/`. They should also go.

PR #327, opened 2025-10-01
Remove projection/content

ostcar:
This is a cleanup for the time, when the projector-service is released and the field projection/content is not be calculated by the autoupdate anymore.

With the new vote service, there is no other field, that is calculated by the autoupdate service. So the functionality can be removed. This also fixes the ambiguous of the term "calculated".

This PR should only be merged, when we are sure, that the projector service is release before or together with rel-db.

If the projector service is released before rel-db, then there is no need to backport this PR on main, since it is only a cleanup PR. There is no harm, that `projector/content` exists together with the new projector-service

### DIFF
--- a/dev/sql/schema_relational.sql
+++ b/dev/sql/schema_relational.sql
@@ -1,7 +1,7 @@
 
 -- schema_relational.sql for initial database setup OpenSlides
 -- Code generated. DO NOT EDIT.
--- MODELS_YML_CHECKSUM = '45a8410dde8705474ea7e4c08430f1df'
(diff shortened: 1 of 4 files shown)


bastianjoel (approved):

hjanott:
We still have comments on `organization/vote_decrypt_public_main_key` and `poll/calculated`. These will be removed in a PR for the new vote service?

hjanott (changes requested):
There is a problem with `test_data.sql` or the schema itself.
```
psql:../sql/test_data.sql:104: ERROR:  column "type" of relation "poll_t" does not exist
```

PR #333, opened 2025-10-14
[rel-db] Add version table

hjanott:


luisa-beerboom (approved):
LGTM, but it seems you have some main-branch commits in here. Idk your policy for keeping the history of the feature branch consistent. I just wanted to mention it because it seems dangerous to me.

PR #341, opened 2025-11-04
New way to vote vor a list of candidates or other optons

ostcar:
Its for polls, where there are many options, but there should be one Yes-No-Abstain-Vote for all options.

In the past, we had this case to vote for a candidate list. The new system would also to vote for a list of text-options.

In the new system, a Yes-No-Abstain-Vote is called an approval-vote. It has its own config. Before this PR, this config did not have a reference to `poll_config_option`. This PR creates an optional reference. For normal motions-polls, there will be no `poll_config_option`. But for candidate-lists, there will be many. The client then shows one `yes-no-abstain`-formular for all of the options.


vkrasnovyd (changes requested):
LGTM and fits into the new voting-related models structure well.

3 back relations have to be removed:
- user/poll_candidate_ids
- meeting/poll_candidate_list_ids
- meeting/poll_candidate_ids

vkrasnovyd (approved):

PR #372, opened 2026-01-17
Changes for new vote service

ostcar:


permission.yml
```diff
@@ -10,12 +10,17 @@ agenda_item:
     can_manage:
         can_see_internal:
             can_see:
+    can_manage_polls: # Can create, start and stop topic polls
+        can_see_polls: # Can see unpublished poll results for topics
+            can_see_internal:
```
vkrasnovyd: Why is agenda_item a special case and has an extra permission level `can_see_internal`? Is it really used in the system? If yes, please add a comment what it is needed for.

ostcar: This has nothing to do with poll and is not new. Its the same permission as in line 11.

For the hierarchical model of our permissions, it is necessary to add with twice (the same as `can_see`).

PR #377, opened 2026-01-24
[rel-DB] Fix check_not_null_for_n_m()

vkrasnovyd:
Fixed error when the trigger was raising error on attempt to delete instance of `own_collection` (meeting_user).

### DIFF
--- a/dev/src/generate_sql_schema.py
+++ b/dev/src/generate_sql_schema.py
@@ -1146,7 +1146,7 @@ class Helper:
                     own_id := NEW.id;
                 ELSE
                     own_id := hstore(OLD) -> ${foreign_column};
-                    EXECUTE format('SELECT 1 FROM %I WHERE "id" = %L', ${own_table}, own_id) INTO counted;
+                    EXECUTE format('SELECT 1 FROM %I WHERE "id" = %L', ${query_relation}, own_id) INTO counted;
                     IF (counted IS NULL) THEN
                         -- if the earlier referenced row was deleted (in the same transaction) we can quit.
                         RETURN NULL;


dev/src/generate_sql_schema.py
```diff
@@ -1050,6 +1051,7 @@ class Helper:
         intermediate_table_foreign_key TEXT := TG_ARGV[4];
         foreign_collection TEXT := TG_ARGV[5];
         foreign_column TEXT := TG_ARGV[6];
+        own_table TEXT := TG_ARGV[7];
```
hjanott: I think we should change `own_collection` to `own_table` and do something like this in the error message where it was already in use.
`SUBSTRING(own_table FOR LENGTH(own_table) - 2)`

dev/src/generate_sql_schema.py
```diff
@@ -646,10 +646,10 @@ def get_trigger_check_not_null_for_1_n(
     def get_trigger_check_not_null_for_n_m(
         cls, own_table_field: TableFieldType, foreign_table_field: TableFieldType
     ) -> str:
-        own_table = own_table_field.table
+        own_collection = own_table_field.table
```
hjanott: Is it really the collection stored in the table attribute?

dev/src/generate_sql_schema.py
```diff
@@ -634,10 +634,10 @@ def get_trigger_check_not_null_for_1_n(
             f"""
             -- definition trigger not null for {own_table}.{own_column} against {foreign_table_t}.{foreign_column}
             CREATE CONSTRAINT TRIGGER {HelperGetNames.get_not_null_rel_list_insert_trigger_name(own_table, own_column)} AFTER INSERT ON {own_table_t} INITIALLY DEFERRED
-            FOR EACH ROW EXECUTE FUNCTION check_not_null_for_1_n('{own_table}', '{own_column}', '');
+            FOR EACH ROW EXECUTE FUNCTION check_not_null_for_1_n('{own_table_t}', '{own_column}', '{foreign_table_t}', '{foreign_column}');
```
hjanott: If this function really receives the collection/view names we should probably rename the parameters.

dev/src/generate_sql_schema.py
```diff
@@ -972,112 +972,197 @@ class Helper:
...
+                    IF (counted IS NULL) THEN
+                        -- if the earlier referenced row was deleted (in the same transaction) we can quit.
+                        RETURN NULL;
+                    END IF;
                 END IF;
-            END IF;
 
-            IF (foreign_id IS NOT NULL) THEN
-                EXECUTE format('SELECT {field_check} FROM %I WHERE id = %s', column_name, table_name, foreign_id) INTO counted;
+                own_collection := SUBSTRING(own_table FOR LENGTH(own_table) - 2);
```
hjanott: Move this to line 999.

dev/src/generate_sql_schema.py
```diff
@@ -972,112 +1133,44 @@ class Helper:
...
+            ${docstring}
+            DECLARE
+            ${parameters_declaration}
+            BEGIN
+                IF (TG_OP = 'INSERT') THEN
+                    -- in case of INSERT the view is checked on itself so the own id is applicable
+                    own_id := NEW.id;
+                ELSE
+                    own_id := hstore(OLD) -> ${foreign_column};
+                    EXECUTE format('SELECT 1 FROM %I WHERE "id" = %L', ${own_table}, own_id) INTO counted;
```
```suggestion
                    EXECUTE format('SELECT 1 FROM %I WHERE "id" = %L', ${query_relation}, own_id) INTO counted;
```

PR #380, opened 2026-02-03
Improve performance of user.committee_ids SQL

rrenkert:


rrenkert:
close in favor of https://github.com/OpenSlides/openslides-meta/pull/381

PR #384, opened 2026-02-12
Move Makefile

Janmtbehrens:
Moving Makefile from /dev into root, for faster access to meta maketargets when debugging


peb-adr (changes requested):

peb-adr (approved):

ostcar:
@Janmtbehrens Can this be merged? Would be a nice change

Janmtbehrens:
@ostcar My latest info is, that this PR should wait until after relDB has been fully rolled out.

ostcar:
@Janmtbehrens what is the status of this? relDB is fully rolled out

Janmtbehrens:
@ostcar the problem right now is that other services like backend reference the moved meta Makefile. These services need a PR as well, but the exact way of ordering their merges so that no invalid state occurs is complicated. We'll have a discussion about this with people from `backend` and then return here

PR #393, opened 2026-02-21
Move shared components to helper_get_names

vkrasnovyd:
Reason: importing from `validate.py` and `join_models_yml.py` into the backend caused test coverage to drop below 96%.
Since the imported values are used mainly for ` InternalHelper.read_models_yml`, it is reasonable to move them to `helper_get_names.py`.

Additional change: `join_models_yml` is now executed in the module mode to correctly resolve relative import from `helper_get_names`.

### DIFF
--- a/dev/requirements.txt
+++ b/dev/requirements.txt
@@ -1,5 +1,5 @@
 autoflake==2.3.1
-black==25.12.0
+black==26.1.0
 flake8==7.3.0
 isort==7.0.0
 sqlfluff==3.4.2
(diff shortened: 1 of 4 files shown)


dev/src/validate.py
```diff
@@ -21,7 +21,6 @@
 COLOR_REGEX = re.compile(r"^#[0-9a-f]{6}$")
 
 ROOT = os.path.join(os.path.dirname(os.path.abspath(__file__)), "..", "..")
-DEFAULT_COLLECTION_META = os.path.join(ROOT, "collection-meta.yml")
 DEFAULT_COLLECTIONS_DIR = os.path.join(ROOT, "collections")
```
hjanott: How about importing this from `helper_get_names`?

vkrasnovyd: Good suggestion. Done. Imported also KEYSEPARATOR.

PR #402, opened 2026-02-27
Generate UNIQUE constraint in schema generation

### LINKED ISSUE #401: Generate UNIQUE constraint in schema generation
Fields like username should be ensured to be unique on database level.

hjanott:
add two ideas on how to deal with special cases of unique relations to yml files
Also implemented a few functions that could help.
Closes #401


dev/src/generate_sql_schema.py
```diff
@@ -1164,7 +1169,7 @@ class Helper:
             ${content_field_indices}
         """))
     GM_FOREIGN_TABLE_LINE_TEMPLATE = string.Template(
-        "    ${gm_content_field} integer GENERATED ALWAYS AS (CASE WHEN split_part(${own_table_column}, '/', 1) = '${foreign_view_name}' THEN cast(split_part(${own_table_column}, '/', 2) AS INTEGER) ELSE null END) STORED CONSTRAINT ${fk_name} REFERENCES ${foreign_table_name}(id) ON DELETE CASCADE INITIALLY DEFERRED,"
+        "    ${gm_content_field} integer GENERATED ALWAYS AS (CASE WHEN split_part(${own_table_column}, '/', 1) = '${foreign_view_name}' THEN cast(split_part(${own_table_column}, '/', 2) AS INTEGER) ELSE null END) STORED CONSTRAINT ${fk_name} REFERENCES ${foreign_table_name}(id) ON DELETE CASCADE INITIALLY DEFERRED${unique},"
```
vkrasnovyd: No need to check every split value: unique `own_table_column` should be enough.

dev/src/generate_sql_schema.py
vkrasnovyd: No need for separate unique constraints. And unique together constraint is already defined (line 1165).

Issue #294, opened 2025-07-31
[rel-db] Add constraint for the `equal_fields`


vkrasnovyd:
Validations related to the `equal_fields` are currently run in the actions. It should also be checked on the database level.

PR #407, opened 2026-03-04
Implement equal fields constraint in schema

luisa-beerboom:
See OpenSlides/openslides-backend#3379

TODOs aside from _The Obvious™_
- [x] Look into constraint triggers
- [x] What about constant models.yml field def? Doesn't seem to be handled in any way?
- [x] generate_sql_schema fails on option/content_object_id, bc it tries to get meeting_id of _user_
- [x] Changed some stuff about the schema generation that may impact generated field names. perhaps ask @peb-adr how to potentially deal with that or change it back -> Looked at it: Apparently it was an edge case that wasn't present in our actual code to begin with.
- [x] CI fails
- [x] refactor and remove the debugging commands (all those print statements, etc)
- [x] cleanup

NOTES:
- Possibly the option/content_object_id meeting_id-equal_fields might have not been previously checked for users


dev/src/generate_sql_schema.py
```diff
@@ -621,9 +636,26 @@ def get_relation_list_type(
         if state != FieldSqlErrorType.ERROR:
             if primary:
                 if foreign_table_field.field_def.get("type") == "relation-list":
-                    nm_table_name, value = Helper.get_nm_table_for_n_m_relation_lists(
+                    (
+                        nm_table_name,
+                        value,
```
hjanott: Let's have a more speaking name for this variable. `definition_text`?

dev/src/generate_sql_schema.py
```diff
@@ -882,6 +914,348 @@ def get_trigger_check_unique_ids_pair(
...
+        equal_fields = table_field.field_def.get("equal_fields")
+        if isinstance(equal_fields, list):
+            return equal_fields
+        elif isinstance(equal_fields, str):
+            return [equal_fields]
+        elif not equal_fields:
+            return []
+        else:
+            raise Exception(
+                f"Could not write equal_fields trigger for {table_field.column}: Unknown setting."
```
```suggestion
                f"Invalid equal_fields for {table_field.column}: Unknown setting."
```

hjanott: Such things should also or mainly be checked in the validate.py.

dev/sql/schema_relational.sql
```diff
@@ -212,6 +212,387 @@ BEGIN
...
+        intermediate_table := TG_ARGV[i];
+        own_table_reference := TG_ARGV[i+1];
+        own_table := TG_ARGV[i+2];
+        foreign_table_reference := TG_ARGV[i+3];
+        foreign_table := TG_ARGV[i+4];
+        check_column := TG_ARGV[i+5];
+        ref_column := TG_ARGV[i+6];
+
+        own_id = NEW.id;
+        FOR r in EXECUTE format('
```
hjanott:
```suggestion
        FOR row in EXECUTE format('
```

dev/sql/schema_relational.sql
```diff
@@ -212,6 +212,387 @@ BEGIN
...
+$check_equals_trigger$ LANGUAGE plpgsql;
+
+-- expects in this order:
+-- * intermediate table name,
+-- * column referencing calling table in intermediate table
+-- * calling table name
+-- * column referencing other table in intermediate table
+-- * other table name
+-- * field that is supposed to be equal
+-- * pseudofield for which the check was triggered
```
hjanott: Pseudo fields means column generated for generic relations. Please explain this term.

dev/sql/schema_relational.sql
```diff
@@ -212,6 +212,387 @@ BEGIN
...
+    intermediate_table TEXT;
+    own_id INTEGER;
+    own_val TEXT;
+    own_table_reference TEXT;
+    own_table TEXT;
+    i INTEGER := 0;
+BEGIN
+
+    WHILE i < TG_NARGS LOOP
+        intermediate_table := TG_ARGV[i];
```
hjanott: unused

dev/sql/schema_relational.sql
```diff
@@ -212,6 +212,387 @@ BEGIN
...
+            own_val,
+            foreign_table,
+            foreign_id,
+            foreign_val
+        );
+
+        i := i + 7;
+    END LOOP;
+
+    RETURN NULL;  -- AFTER TRIGGER needs no return
```
hjanott: After triggers always need return but the return value is ignored. Please also change the comment where you copied it from and copied it to.

dev/sql/schema_relational.sql
```diff
@@ -212,6 +212,387 @@ BEGIN
...
+DECLARE
+    ref_column TEXT;
+    check_column TEXT;
+    foreign_table_reference TEXT;
+    foreign_table TEXT;
+    foreign_id INTEGER;
+    foreign_val TEXT;
+    intermediate_table TEXT;
+    own_id INTEGER;
+    own_val TEXT;
```
hjanott: I would name these `_val`-variables `_equal_val` to give it more differentiating meaning. There are also cases in other functions.

dev/sql/schema_relational.sql
```diff
@@ -212,6 +212,387 @@ BEGIN
...
+        own_table := TG_ARGV[i];
+        ref_column := TG_ARGV[i+1];
+        EXECUTE format(
+            'SELECT ($1).id, ($1).meeting_id, ($1).%I',
+            ref_column
+        ) INTO own_id, own_val, user_id USING NEW;
+        IF user_id IS NOT NULL THEN
+            EXECUTE format(
+                'SELECT meeting_id
+                FROM meeting_user_t
```
hjanott: Hardcoding tables will break in migrations.

dev/sql/schema_relational.sql
```diff
@@ -212,6 +212,387 @@ BEGIN
 END;
 $read_only_trigger$ LANGUAGE plpgsql;
 
+CREATE OR REPLACE FUNCTION raise_equality_exception(check_column TEXT, ref_column TEXT, own_table TEXT, own_id INTEGER, own_val TEXT, foreign_table TEXT, foreign_id INTEGER, foreign_val TEXT)
```
hjanott:
```suggestion
CREATE OR REPLACE FUNCTION raise_equality_exception_conditionally(check_column TEXT, ref_column TEXT, own_collection TEXT, own_id INTEGER, own_val TEXT, foreign_collection TEXT, foreign_id INTEGER, foreign_val TEXT)
```

dev/sql/schema_relational.sql
```diff
@@ -212,6 +212,387 @@ BEGIN
...
+    RETURN NULL;  -- AFTER TRIGGER needs no return
+END;
+$check_equals_intermediate_trigger$ LANGUAGE plpgsql;
+
+-- expects in this order:
+-- * own table name (i.e. name of table that isn't user),
+-- * user relation field in said table for which the check was triggered
+-- checks if meeting_id of NEW is equal to meeting_id of connected user,
+-- which is grandfathered in from whichever meeting_user connects that user to that meeting.
+CREATE OR REPLACE FUNCTION check_equals_meeting_id_for_user()
```
hjanott, line 537: Could the multi trigger be used for this case by adjusting the parameters? Just a hunch of what these actually do. I don't like that we split into so many functions.

dev/sql/schema_relational.sql
```diff
@@ -212,6 +212,387 @@ BEGIN
...
+    RETURN NULL;  -- AFTER TRIGGER needs no return
+END;
+$check_equals_meeting_id_for_user_trigger$ LANGUAGE plpgsql;
+
+-- called on meeting_user delete.
+-- expects in this order:
+-- * own table name (i.e. name of table that isn't user),
+-- * user relation field in said table for which the check was triggered
+-- Checks if the other table has any row with the same meeting_id pointing to that user.
+CREATE OR REPLACE FUNCTION check_equals_meeting_id_user_on_meeting_user_delete()
```
hjanott, line 590: I hope the new vote service will reference meeting_users instead of users as I don't like so many specialized functions.

dev/sql/schema_relational.sql
```diff
@@ -3596,6 +3973,599 @@ CREATE TRIGGER tr_log_vote_t_meeting_id AFTER INSERT OR UPDATE OF meeting_id OR
...
+FOR EACH ROW EXECUTE FUNCTION check_equals('projection', 'projector', 'preview_projector_id', 'meeting_id', TRUE);
+
+
+CREATE CONSTRAINT TRIGGER equal_meeting_id_on_projection_t_history_projector_id AFTER INSERT OR UPDATE OF history_projector_id, meeting_id ON projection_t INITIALLY DEFERRED
+FOR EACH ROW EXECUTE FUNCTION check_equals('projection', 'projector', 'history_projector_id', 'meeting_id', FALSE);
+CREATE CONSTRAINT TRIGGER equal_meeting_id_on_projector_t_history_projection_ids AFTER INSERT ON projector_t INITIALLY DEFERRED
+FOR EACH ROW EXECUTE FUNCTION check_equals('projection', 'projector', 'history_projector_id', 'meeting_id', TRUE);
+
+
+CREATE CONSTRAINT TRIGGER equal_meeting_id_on_projection_t_content_object_id_meeting_id AFTER INSERT ON projection_t INITIALLY DEFERRED
```
hjanott, line 5242: Also need this on update of `content_object_id_meeting_id` and `meeting_id`.

dev/src/generate_sql_schema.py
```diff
@@ -882,6 +914,348 @@ def get_trigger_check_unique_ids_pair(
...
+                    CREATE CONSTRAINT TRIGGER {own_trigger_name} AFTER {own_event_str} ON {own_table} INITIALLY DEFERRED
+                    FOR EACH ROW EXECUTE FUNCTION check_equals_meeting_id_for_user('{own_table_field.table}', '{specified_relation_field}');
+                    CREATE CONSTRAINT TRIGGER {back_trigger_name} AFTER DELETE ON meeting_user_t INITIALLY DEFERRED
+                    FOR EACH ROW EXECUTE FUNCTION check_equals_meeting_id_user_on_meeting_user_delete('{own_table_field.table}', '{specified_relation_field}');
+
+                """)
+            elif foreign_table_field.table == "meeting" and equal_field == "meeting_id":
+                sql += dedent(f"""
+                    CREATE CONSTRAINT TRIGGER {own_trigger_name} AFTER {own_event_str} ON {own_table} INITIALLY DEFERRED
+                    FOR EACH ROW EXECUTE FUNCTION check_equals_meeting_id_for_meeting('{own_table_field.table}', '{specified_relation_field}');
```
hjanott, line 1127: I have a strong feeling that something like this would work. Would probably need an additional flag for using `OLD`.
```suggestion
                    FOR EACH ROW EXECUTE FUNCTION check_equals('{own_table_field.table}', '{own_table_field.table}', '{specified_relation_field}', '{equal_field}');
```

vkrasnovyd (changes requested):

dev/src/generate_sql_schema.py
```diff
@@ -882,6 +914,348 @@ def get_trigger_check_unique_ids_pair(
...
+                        f"Cannot generate equal_fields trigger: user/meeting_id handling not implemented for {table_field.collectionfield}"
+                    )
+                field_def = InternalHelper.get_models("meeting_user", "meeting_id")
+                own_table_field = TableFieldType(
+                    "meeting_user", "meeting_id", field_def
+                )
+                foreign_table_fields: TableFieldType | list[TableFieldType] = (
+                    TableFieldType(collection, field, field_def)
+                )
+                fse_type, primary, b, c = InternalHelper.check_relation_definitions(
```
```suggestion
                fse_type, *_ = InternalHelper.check_relation_definitions(
```

dev/src/generate_sql_schema.py
```diff
@@ -882,6 +914,348 @@ def get_trigger_check_unique_ids_pair(
...
+                                field_def.get("to"), field_def.get("reference")
+                            )
+                        ]
+                    else:
+                        foreign_table_fields = (
+                            InternalHelper.get_definitions_from_foreign_list(
+                                field_def.get("to"), field_def.get("reference")
+                            )
+                        )
+                    fse_type, primary, b, c = InternalHelper.check_relation_definitions(
```
```suggestion
                    fse_type, *_ = InternalHelper.check_relation_definitions(
```

dev/src/generate_sql_schema.py
```diff
@@ -882,6 +914,348 @@ def get_trigger_check_unique_ids_pair(
...
+    def get_equal_field_trigger_config(
+        cls,
+        table_field: TableFieldType,
+        fields: list[TableFieldType | str],
+        fail_on_user_meeting_id: bool = True,
+        fail_on_meeting_meeting_id: bool = True,
+    ) -> tuple[str, bool]:
+        """
+        Checks the configuration of the relation and returns:
+        - The name of the table/view that should be used
```
```suggestion
        - The name of the table that should be used
```

dev/src/generate_sql_schema.py
```diff
@@ -882,6 +914,348 @@ def get_trigger_check_unique_ids_pair(
...
+                foreign_table_field, [equal_field], False, False
+            )
+            full_own_trigger_name = (
+                f"equal_meeting_id_on_{own_table}_{specified_relation_field}"
+            )
+            own_trigger_name = HelperGetNames.get_shortened_name(full_own_trigger_name)
+            if foreign_table_field.table == "user" and equal_field == "meeting_id":
+                back_trigger_name = HelperGetNames.get_shortened_name(
+                    f"{full_own_trigger_name}_back"
+                )
```
hjanott: Please create a `HelperGetNames` method for this with a `max_len` decorator. As @vkrasnovyd mentioned this is necessary in all places where a name is generated.

dev/src/generate_sql_schema.py
```diff
@@ -882,6 +914,348 @@ def get_trigger_check_unique_ids_pair(
...
+        fail_on_user_meeting_id: bool = True,
+        fail_on_meeting_meeting_id: bool = True,
+    ) -> tuple[str, bool]:
+        """
+        Checks the configuration of the relation and returns:
+        - The name of the table/view that should be used
+        - If the field can be updated
+        """
+        with_update = False
+        use_view = False
```
hjanott: More of a `fail` or `error` flag.

dev/src/generate_sql_schema.py
```diff
@@ -882,6 +914,348 @@ def get_trigger_check_unique_ids_pair(
...
+        if state != FieldSqlErrorType.FIELD:
+            raise Exception(
+                f"Could not write equal_fields trigger for {table_field.column}: Not supported for FieldSqlErrorType {state}."
+            )
+
+    @classmethod
+    def get_equal_field_trigger_config(
+        cls,
+        table_field: TableFieldType,
+        fields: list[TableFieldType | str],
```
hjanott: Thought. There should be a better way to create these by considering both fields at the same time. not in a loop.

dev/src/generate_sql_schema.py
```diff
@@ -1199,6 +1609,391 @@ class Helper:
...
+
+        -- expects in this order:
+        -- * intermediate table name,
+        -- * column referencing calling table in intermediate table
+        -- * calling table name
+        -- * column referencing other table in intermediate table
+        -- * other table name
+        -- * field that is supposed to be equal
+        -- * pseudofield for which the check was triggered
+        CREATE OR REPLACE FUNCTION check_equals_multi()
```
hjanott, line 1654: Please think how you can map `check_equals_intermediate` into this. You have `    IF (TG_OP = 'UPDATE') THEN` to your disposal.

dev/src/generate_sql_schema.py
```diff
@@ -1199,6 +1609,391 @@ class Helper:
...
+                        JOIN %I c ON b.%I = c.id
+                    WHERE a.id = $1',
+                    check_column,
+                    check_column,
+                    own_table,
+                    intermediate_table,
+                    own_table_reference,
+                    foreign_table,
+                    foreign_table_reference
+                ) USING own_id LOOP
```
hjanott: Using own_id? Why that?

dev/src/generate_sql_schema.py
```diff
@@ -1199,6 +1609,391 @@ class Helper:
...
+
+            WHILE i < TG_NARGS LOOP
+                own_table := TG_ARGV[i];
+                ref_column := TG_ARGV[i+1];
+                muser_table_identifier := TG_ARGV[i+2];
+                EXECUTE format(
+                    'SELECT ($1).id, ($1).meeting_id, ($1).%I',
+                    ref_column
+                ) INTO own_id, own_val, user_id USING NEW;
+                IF user_id IS NOT NULL THEN
```
hjanott: Is it possible that a similar check is missing in `check_equals`?

dev/src/generate_sql_schema.py
hjanott, line 2395: I know this is not part of your PR but please write a function for this in `HelperGetNames`. Same reason as with other names this should not exceed 63 characters.

dev/src/generate_sql_schema.py
```diff
@@ -882,6 +914,348 @@ def get_trigger_check_unique_ids_pair(
...
+        foreign_table_field: TableFieldType,
+        state: FieldSqlErrorType,
+    ) -> str:
+        cls.equal_fields_state_check(state, own_table_field)
+        equal_fields = list(
+            {
+                *cls.get_equal_fields(own_table_field),
+                *cls.get_equal_fields(foreign_table_field),
+            }
+        )
```
hjanott: This usage is of `get_equal_fields` repetitive and the only kind. Creating a set from `equal_fields` of both fields in `get_equal_fields` would be possible for list and string, can be returned unified and can be iterated over in the loop the same way.

dev/src/generate_sql_schema.py
```diff
@@ -621,9 +636,26 @@ def get_relation_list_type(
...
+                        nm_table_name,
+                        value,
+                        own_intermediate_field,
+                        foreign_intermediate_field,
+                    ) = Helper.get_nm_table_for_n_m_relation_lists(
                         own_table_field, foreign_table_field
                     )
+                    if own_table_field.field_def.get(
+                        "equal_fields"
+                    ) or foreign_table_field.field_def.get("equal_fields"):
```
hjanott: Could use `get_equal_fields` function.

dev/src/generate_sql_schema.py
```diff
@@ -882,6 +914,348 @@ def get_trigger_check_unique_ids_pair(
...
+                foreign_table_field, [equal_field], False
+            )
+            if (
+                "reference" in own_table_field.field_def
+                and "reference" in foreign_table_field.field_def
+                and foreign_table_field.field_def.get("type") in ["relation"]
+            ):
+                raise Exception(
+                    f"Cannot generate equal_fields triggers for {own_table_field.collectionfield} and {foreign_table_field.collectionfield}: Both have reference set."
+                )
```
hjanott, line 1013: Such checks should reside in validate.py in my opinion.

dev/src/generate_sql_schema.py
```diff
@@ -918,6 +1293,15 @@ def get_generic_relation_type(
                     own_table_field.column,
                     foreign_table_field,
                 )
+                if own_table_field.field_def.get(
+                    "equal_fields"
+                ) or foreign_table_field.field_def.get("equal_fields"):
```
hjanott: Could also be a place to use `get_equal_fields`.

dev/src/generate_sql_schema.py
```diff
@@ -973,6 +1364,25 @@ def get_generic_relation_list_type(
...
                         f"Tried to create gm_table '{gm_foreign_table}' twice"
                     )
+                equal_fields_text = ""
+                for (
+                    foreign_intermediate_field,
+                    foreign_table_field,
+                ) in foreign_intermediate_field_foreign_table_field.items():
+                    if own_table_field.field_def.get(
+                        "equal_fields"
+                    ) or foreign_table_field.field_def.get("equal_fields"):
```
hjanott: Could also be a place to use `get_equal_fields`.

dev/sql/schema_relational.sql
```diff
@@ -212,6 +212,391 @@ BEGIN
...
+BEGIN
+    IF foreign_id IS NOT NULL AND own_id IS NOT NULL THEN
+        IF foreign_val IS DISTINCT FROM own_val THEN
+            foreign_fqid := foreign_table || '/' || foreign_id;
+            IF check_column = 'meeting_id' THEN
+                RAISE EXCEPTION 'The following models do not belong to meeting %: [''%'']', own_val, foreign_fqid;
+            END IF;
+            foreign_fqid := foreign_fqid  || '/' || check_column;
+            own_fqid := own_table || '/' || own_id || '/' || check_column;
+            RAISE EXCEPTION 'The relation % requires the following fields to be equal:% %: % % %: %', ref_column, chr(10), own_fqid, own_val, chr(10), foreign_fqid, foreign_val;
```
hjanott: I would like to combine these two error messages. Something along the lines:
`                RAISE EXCEPTION 'The following models do not belong to % %: [''%'']' \nThe following fields need to be equal:% %: % % %: %', own_collection, own_val, foreign_fqid, ref_column, chr(10), own_fqid, own_val, chr(10), foreign_fqid, foreign_val;`

dev/sql/schema_relational.sql
```diff
@@ -212,6 +212,391 @@ BEGIN
...
+
+        i := i + 2;
+    END LOOP;
+
+    RETURN NULL;  -- AFTER TRIGGER needs no return
+END;
+$check_equals_meeting_id_user_on_meeting_user_delete_trigger$ LANGUAGE plpgsql;
+
+CREATE OR REPLACE FUNCTION check_equals_meeting_id_for_meeting()
+RETURNS trigger AS $check_equals_meeting_id_for_meeting$
```
hjanott, line 638: Did I mention that I don't like so many specialized functions? I know I know. It's not that easy. But if I compare this to `check_equals` it's pretty much the same. Probably you just need to pass the parameters to `check_equals` in a specialized way.

hjanott (changes requested):
Sorry for the extensive review. This PR is really complex and not easy to delve into. I'm cutting this short. So I may have missed some things more important than what is mentioned here.

dev/src/generate_sql_schema.py
```diff
@@ -882,6 +914,348 @@ def get_trigger_check_unique_ids_pair(
...
+                "reference" in own_table_field.field_def
+                and "reference" in foreign_table_field.field_def
+                and foreign_table_field.field_def.get("type") in ["relation"]
+            ):
+                raise Exception(
+                    f"Cannot generate equal_fields triggers for {own_table_field.collectionfield} and {foreign_table_field.collectionfield}: Both have reference set."
+                )
+            full_own_trigger_name = (
+                f"equal_{equal_field}_on_{own_table}_{own_table_field.column}"
+            )
```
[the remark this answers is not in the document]
luisa-beerboom: Did that in the merge

dev/src/generate_sql_schema.py
luisa-beerboom: Which one?

dev/src/generate_sql_schema.py
luisa-beerboom: Changed message and added a check to the validate.py

dev/sql/schema_relational.sql
luisa-beerboom: Done

dev/sql/schema_relational.sql
luisa-beerboom: I don't understand what you want exactly, but I changed the description

dev/sql/schema_relational.sql
luisa-beerboom: I don't see anything that is unused here, what are you referring to?

dev/sql/schema_relational.sql
luisa-beerboom: Changed it to `-- returning NULL because AFTER TRIGGER return value is ignored`

dev/sql/schema_relational.sql
luisa-beerboom: Done

dev/sql/schema_relational.sql
luisa-beerboom: I believe I fixed that already when fixing the migration code in the backend.

dev/sql/schema_relational.sql
luisa-beerboom: Done

dev/sql/schema_relational.sql
luisa-beerboom: Technically everything is possible, but I think the multi function is already complex enough.

dev/sql/schema_relational.sql
luisa-beerboom: `projection/meeting_id` and `projection/content_object_id` are both constant and therefore cannot be updated. That's why the trigger doesn't generate to listen to their updates.

dev/src/generate_sql_schema.py
luisa-beerboom: It doesn't work as is.
It would be theoretically possible to rewrite check_equals with special flags and special handling to deal with it, however:
1. That would only further complicate the already complex `check_equals` function
2. For nothing but a tiny little special case, which I would hope will someday vanish from our codebase entirely
3. `check_equals_meeting_id_for_meeting` doesn't even have significant functional overlap

dev/src/generate_sql_schema.py
luisa-beerboom: Done

dev/src/generate_sql_schema.py
luisa-beerboom: Done

dev/src/generate_sql_schema.py
luisa-beerboom: Done

dev/src/generate_sql_schema.py
luisa-beerboom: Done when fixing the main merge

dev/src/generate_sql_schema.py
luisa-beerboom: Renamed to `view_field_error`

dev/src/generate_sql_schema.py
luisa-beerboom: "both fields at the same time"
What two fields and what loop are you referring to and what are "these"?

Please describe your thoughts in a more detailed manner.

dev/src/generate_sql_schema.py
luisa-beerboom: I built separate functions  for these cases since they get their data and process it through completely different methods, due to looking at the problem from completely different functional parts of the relation.
I firmly believe, merging them would only be possible by doing something like putting a giant if clause in the resulting method and putting the `check_equals_multi` code and the `check_equals_intermediate` code on either side.
I much prefer having shorter, less complex and more readable functions, rather than giant, complex, unreadable ones where you need to constantly consider what case the function would be handling right now based on the specific configuration of countless boolean flags.

dev/src/generate_sql_schema.py
luisa-beerboom: Makes no difference functionally, but I moved it to the replacements list, to have uniform syntax.

dev/src/generate_sql_schema.py
luisa-beerboom: The check technically stopped the select below from uselessly being called, but changed nothing about the functionality. I removed it for easier reading.

dev/src/generate_sql_schema.py
luisa-beerboom: Done begrudgingly.

dev/src/generate_sql_schema.py
luisa-beerboom: Done

dev/src/generate_sql_schema.py
luisa-beerboom: Done

dev/src/generate_sql_schema.py
hjanott: value

dev/sql/schema_relational.sql
hjanott: `intermediate_table`
Please also don't forget to remove the parameters and update the increment then.

dev/sql/schema_relational.sql
hjanott: Now that I look at it. This should also be changed in the calling functions.

dev/src/generate_sql_schema.py
hjanott: I love how that works now.

dev/src/generate_sql_schema.py
luisa-beerboom: Removed the error code from the schema generator and reconstructed them in validate.py

dev/src/generate_sql_schema.py
luisa-beerboom: Done

dev/src/generate_sql_schema.py
luisa-beerboom: Done

dev/src/generate_sql_schema.py
hjanott: I clearly meant the up to two fields passed in fields. Whatever else I was thinking, it probably didn't make sense. I like how clear this function became however.

dev/sql/schema_relational.sql
luisa-beerboom: I don't think hanging the second onto the end of the first makes sense.
The second error message would make the first obsolete since it would imply the same information and more.

Considering that this PR is intended to to fix current functionality while changing it as little as possible. Given that the meeting-specific message is by far more commonly used, this would be a somewhat major change, so I'd prefer not to do it here.

It would probably be more productive to discuss this stuff separately with the affected parties (that includes the client devs), decide what messages would be appropriate then and then _potentially_ change these things in their own PR

dev/sql/schema_relational.sql
luisa-beerboom: 1. We discussed this.
2. I don't want to overcomplicate `check_equals` with a bunch of edge-case code that'll make it even less readable
3. I don't want to overcomplicate `check_equals` with a bunch of edge-case code that'll have to be checked everytime, but only be relevant in one single tiny edge case.
4. As far as I'm concerned this edge case shouldn't even exist, so maybe we should rather seek to find a better solution than `equal_fields` to the underlying problem, implement it and then _delete this function,_ once reldb is running properly.

dev/src/generate_sql_schema.py
luisa-beerboom: I changed it, but I'll point out, I do not like being made to refactor/cleanup code I didn't write or to change/rethink _variable names I didn't invent._

dev/sql/schema_relational.sql
luisa-beerboom: Ah. I was looking at the wrong function it seems.
Changed.

dev/sql/schema_relational.sql
luisa-beerboom: Done

dev/sql/schema_relational.sql
hjanott: I don't see the change. There seem to be multiple places where this description was used.
What I think this is, is the column that gets generated for generic relations to create foreign key relations and the like but is no field in the models definition. We don't have an official term for this yet. Pseudo field may be fitting but should be explained somewhere. Maybe here.

vkrasnovyd:
Resolves https://github.com/OpenSlides/openslides-meta/issues/294

dev/src/validate.py
```diff
@@ -442,21 +442,90 @@ def check_relation(
...
+                        and all(isinstance(val, str) for val in equal_fields)
+                    )
+                ):
+                    self.errors.append(
+                        f"'equal_fields' of {collection_field} is not valid (must be string or list of strings)."
+                    )
+                    return
+        to_eq_fields = self.listify(to_field.get("equal_fields"))
+        from_eq_fields = self.listify(from_field.get("equal_fields"))
+        joined_eq_fields = {*to_eq_fields, *from_eq_fields}
```
hjanott: Why not create and use `setify` instead? You could even use the `get_equal_fields` from `generat_sql_schema.py`. I really liked that one.

dev/sql/schema_relational.sql
hjanott: I created an issue for that. #458

dev/src/generate_sql_schema.py
hjanott: I created an issue for that. #458

dev/sql/schema_relational.sql
hjanott: I created an issue for that. #458

dev/sql/schema_relational.sql
hjanott: I created an issue for that. #458

dev/src/generate_sql_schema.py
hjanott: #460 is the place for this now.

vkrasnovyd (approved):
Checked only the updated trigger names generation (the change that I've requested). That's exactly what I wanted to see.

dev/sql/schema_relational.sql
luisa-beerboom: Seems that I only changed it in another place ('^_^)
Changing the description for check_equals_multi as well.

dev/src/validate.py
luisa-beerboom: Turned it to `setify_equal_fields` and then used union. Made it a bit shorter that way.

hjanott (approved):
Thanks for the hard work. Looks ready for me now.

PR #426, opened 2026-03-20
Shorten constraints names

vkrasnovyd:


hjanott (approved):
So much better like this. Can be merged after unifying `max_length` and `get_shortened_name`.

PR #447, opened 2026-04-02
Add constraints names

### LINKED ISSUE #437: Create names for all constraints
Currently fields like color have check constraints that are not named. Also other constraints like NOT NULL are not named by us. To keep this in line we should generate names for all of these.

vkrasnovyd:
Resolves #437


dev/src/generate_sql_schema.py
```diff
@@ -1244,17 +1245,30 @@ class Helper:
...
+                    CONSTRAINT ${required_constraint_name_1} NOT NULL
+                    CONSTRAINT ${fk_name_1} REFERENCES ${table1} (id)
+                    ON DELETE CASCADE
+                    INITIALLY DEFERRED,
+                ${field2} integer
+                    CONSTRAINT ${required_constraint_name_2} NOT NULL
+                    CONSTRAINT ${fk_name_2} REFERENCES ${table2} (id)
+                    ON DELETE CASCADE
+                    INITIALLY DEFERRED,
+                CONSTRAINT ${pk_constraint_name} PRIMARY KEY (${list_of_keys})
```
hjanott: This got so long. Maybe its time to create a template string just for the fields. Especially since they are basically the same.

hjanott (changes requested):

hjanott (approved):

PR #453, opened 2026-04-07
[rel-db] Add unique_together constraint for MeetingMediafile

### LINKED ISSUE #449: [rel-db] `unique_together` constraint for `MeetingMediafile`
The `mediafile_id` and `meeting_id` fields of a `MeetingMediafile` should be unique together.
This is currently only handled [by the backend](https://github.com/OpenSlides/openslides-backend/blob/4a4857b/openslides_backend/action/actions/meeting_mediafile/create.py#L26).

Follow Up to: #412

MoonCubes:
Resolves #449 

Follow Up to #412

### DIFF
--- a/collections/meeting_mediafile.yml
+++ b/collections/meeting_mediafile.yml
@@ -1,3 +1,6 @@
+unique_together:
+- mediafile_id, meeting_id 
+
 fields:
   id:
     type: number
--- a/dev/sql/schema_relational.sql
+++ b/dev/sql/schema_relational.sql
@@ -1,7 +1,7 @@
 
 -- schema_relational.sql for initial database setup OpenSlides
 -- Code generated. DO NOT EDIT.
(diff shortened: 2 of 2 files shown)


hjanott (approved):
Looks good to me. Should be merged after backend PR.

MoonCubes:
> Should be merged after backend PR.

Shouldn't this be merged first?

hjanott:
> Shouldn't this be merged first?

That would be the intuitive way to do things. But to make the tests run in the backend (https://github.com/OpenSlides/openslides-backend/pull/3463) you will have to add this branch as the submodule hash there. Then we can merge the backend. But we will have to update the hash again with the new hash after this got merged to not be pointing to your branch anymore. For that we have an automation the creates pull requests in all services that use the meta.
Basically this process is like this to avoid having to run the backend tests again after merging the automatic meta bumps.

Issue #461, opened 2026-04-14
Constraints based on is_directory value


vkrasnovyd:
`token`, `filesize`, `filename`, `pdf_information`, and `mimetype` are allowed only when `is_directory` == `False`. `filename` is required in this case. Sources:
- https://github.com/OpenSlides/OpenSlides/wiki/Mediafiles
- schema of `mediafile.upload` backend-action

PR #462, opened 2026-04-15
Add unique_together_strict

vkrasnovyd:
Introduces `unique_together_strict` field for collection files. The generated from it unique together constraint has `NULLS NOT DISTINCT`.

For this constraint to work properly, marked `title` as required (the check already exists in the backend).


collections/mediafile.yml
hjanott, line 27: Shouldn't filename be required too? Or is that for another PR?

vkrasnovyd: According to the actions schema, it is allowed and required only when file is not a directory. Issue: https://github.com/OpenSlides/openslides-meta/issues/461

hjanott (approved):
LGTM.

PR #477, opened 2026-04-21
Remove option user equal_fields trigger

luisa-beerboom:
See OpenSlides/openslides-backend#3508

### DIFF
--- a/dev/src/generate_sql_schema.py
+++ b/dev/src/generate_sql_schema.py
@@ -966,7 +966,12 @@ class GenerateCodeBlocks:
             if isinstance(field, TableFieldType):
                 # Assume that these are always primary
                 field_def = field.field_def
-            elif collection == "user" and field == "meeting_id":
+            elif (
+                collection == "user"
+                and field == "meeting_id"
+                and "meeting_id" not in InternalHelper.MODELS["user"]["fields"]
+            ):
+                return "", False
                 field_def = InternalHelper.get_models("meeting_user", "meeting_id")
             elif collection == "meeting" and field == "meeting_id":
                 field_def = None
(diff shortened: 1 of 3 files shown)


dev/src/generate_sql_schema.py
```diff
@@ -966,7 +966,12 @@ def get_equal_field_trigger_config(
...
                 # Assume that these are always primary
                 field_def = field.field_def
-            elif collection == "user" and field == "meeting_id":
+            elif (
+                collection == "user"
+                and field == "meeting_id"
+                and "meeting_id" not in InternalHelper.MODELS["user"]["fields"]
+            ):
+                return "", False
                 field_def = InternalHelper.get_models("meeting_user", "meeting_id")
```
hjanott: Should be dead code with the early return.
```suggestion
```

dev/src/generate_sql_schema.py
```diff
@@ -1110,18 +1112,19 @@ def get_trigger_check_equal_fields_for_g1_x(
...
-                    FOR EACH ROW EXECUTE FUNCTION check_equals_meeting_id_for_user('{own_table_field.table}', '{specified_relation_field}', 'meeting_user_t');
-                    CREATE CONSTRAINT TRIGGER {back_trigger_name} AFTER DELETE ON meeting_user_t INITIALLY DEFERRED
-                    FOR EACH ROW EXECUTE FUNCTION check_equals_meeting_id_user_on_meeting_user_delete('{own_table_field.table}', '{specified_relation_field}');
-
-                """)
-            elif foreign_table_field.table == "meeting" and equal_field == "meeting_id":
+            elif (
+                foreign_table_field.table == "meeting"
+                and equal_field == "meeting_id"
+                and "meeting_id" not in InternalHelper.MODELS["meeting"]["fields"]
```
hjanott, line 1110: Why did you add in all these cases this line? The `MODELS` entry should be constant and never contain `"meeting_id"` for these collections.

hjanott (changes requested):
I'm so happy to see this. However the custom-ness in the generate script still bugs me a little. What are the reasons to not simply remove the equal fields constraint from the options collection definition? It should still generate the constraints for the other two related collections right?

dev/src/generate_sql_schema.py
luisa-beerboom: Just for security reasons, since we can't be sure what kind of whacky features will be requested in future, I thought it better to be safe than sorry.

dev/src/generate_sql_schema.py
luisa-beerboom: Seems I forgot to delete that

luisa-beerboom:
> What are the reasons to not simply remove the equal fields constraint from the options collection definition? It should still generate the constraints for the other two related collections right?

So long as the other two sides have equal_fields set, yes.
Fair point, I will do that and then we can delete the remaining user-related code.

rrenkert:
See https://github.com/OpenSlides/openslides-backend/pull/3513#issuecomment-4305993144
Removed staging-t1 label

hjanott (approved):
Purrfect

PR #479, opened 2026-04-21
Add log triggers for calculated fields

### LINKED ISSUE #445: Accounts: Adding to meeting doesn't work
**Describe the bug**
Adding accounts to a meeting via bulk operation in Accounts menu (B) or as single function in account detail (A) doesn't work. It happens nothing. 

**How to Reproduce**
1. precondition: A meeting is needed
2. open Accounts
3. create two Account Max and Moritz
4. (B) open account Max > three-dot-menu > add to meetings > add the user to a meeting. Nothing happens
5. (A) open Accounts > three-dot-menu > Multiselect > select account Moritz > three-dot-menu > set/remove meeting > select a meeting > add 

**Expected behavior**
Adding accounts via bulk operation or via account detail menu to a meeting must be possible. 

**Additional context**
This works on 4.2.29. It is a problem in the upcoming version 4.3

vkrasnovyd:
Resolves https://github.com/OpenSlides/openslides-meta/issues/445

TODO:
- [x] (needs trigger for constant fields): if both sides of nm-table or all the fields from `define_trigger_on.columns` have `constant: true` + `required: true`, update operation can be excluded from the trigger definition
- [x] Collections validator must check that:
    - [x] If `sql` is defined, `log_triggers_data` is defined too (+ add it to allowed attributes in this case);
    - [x] For `log_collection_id` and `log_value` one of `column` or `sql` is defined;
    - [x] For `define_trigger_on` `table` is defined and ends with '_t';
    - [x] `relational_column` is defined.

### DIFF
--- a/collections/committee.yml
+++ b/collections/committee.yml
@@ -52,9 +52,33 @@ fields:
           UNION
 
           -- Select user_id from home committees
-          SELECT u.id FROM user_t u WHERE u.home_committee_id = c.id
+          SELECT u.id FROM user_t u
+          WHERE u.home_committee_id = c.id
         ) _
       ) AS user_ids
+    log_triggers_data:
+      - define_trigger_on:
+          table: meeting_user_t
+        relational_column: meeting_id
+        log_collection_id:
+          sql: SELECT committee_id FROM meeting_t WHERE id = ($1).meeting_id
+        log_value:
+          column: user_id
+      - define_trigger_on:
+          table: nm_committee_manager_ids_user_t
+        relational_column: committee_id
+        log_collection_id:
+          column: committee_id
+        log_value:
+          column: user_id
+      - define_trigger_on:
+          table: user_t
+          update_columns: home_committee_id
+        relational_column: home_committee_id
+        log_value:
+          column: id
+        log_collection_id:
+          column: home_committee_id
   manager_ids:
     type: relation-list
     to: user/committee_management_ids
--- a/collections/committee.yml
+++ b/collections/committee.yml
@@ -57,29 +57,20 @@ fields:
           WHERE u.home_committee_id = c.id
         ) _
       ) AS user_ids
-    log_triggers_data:
-      - define_trigger_on:
-          table: meeting_user_t
+    log_triggers:
+      - on_table: meeting_user_t
         relational_column: meeting_id
-        log_collection_id:
-          sql: SELECT committee_id FROM meeting_t WHERE id = ($1).meeting_id
-        log_value:
-          column: user_id
-      - define_trigger_on:
-          table: nm_committee_manager_ids_user_t
+        log_collection_id_sql: SELECT committee_id FROM meeting_t WHERE id = ($1).meeting_id
+        log_value_column: user_id
+      - on_table: nm_committee_manager_ids_user_t
         relational_column: committee_id
-        log_collection_id:
-          column: committee_id
-        log_value:
-          column: user_id
-      - define_trigger_on:
-          table: user_t
-          update_columns: home_committee_id
+        log_collection_id_column: committee_id
+        log_value_column: user_id
+      - on_table: user_t
+        on_columns: home_committee_id
         relational_column: home_committee_id
-        log_value:
-          column: id
-        log_collection_id:
-          column: home_committee_id
+        log_collection_id_column: home_committee_id
+        log_value_column: id
   manager_ids:
     type: relation-list
     to: user/committee_management_ids
--- a/collections/committee.yml
+++ b/collections/committee.yml
@@ -59,16 +59,13 @@ fields:
       ) AS user_ids
     log_triggers:
       - on_table: meeting_user_t
-        relational_column: meeting_id
         log_collection_id_sql: SELECT committee_id FROM meeting_t WHERE id = ($1).meeting_id
         log_value_column: user_id
       - on_table: nm_committee_manager_ids_user_t
-        relational_column: committee_id
         log_collection_id_column: committee_id
         log_value_column: user_id
       - on_table: user_t
         on_columns: home_committee_id
-        relational_column: home_committee_id
         log_collection_id_column: home_committee_id
         log_value_column: id
   manager_ids:
--- a/collections/committee.yml
+++ b/collections/committee.yml
@@ -52,7 +52,8 @@ fields:
           UNION
 
           -- Select user_id from home committees
-          SELECT u.id FROM user_t u
+          SELECT u.id
+          FROM user_t u
           WHERE u.home_committee_id = c.id
         ) _
       ) AS user_ids
--- a/collections/user.yml
+++ b/collections/user.yml
@@ -127,6 +127,30 @@ fields:
           WHERE u_hc.home_committee_id IS NOT NULL AND u_hc.id = u.id
         ) AS ci
       ) AS committee_ids
+    log_triggers_data:
+      - define_trigger_on:
+          table: meeting_user_t
+        relational_column: meeting_id
+        log_collection_id:
+          column: user_id
+        log_value:
+          column:
+          sql: SELECT committee_id FROM meeting_t WHERE id = ($1).meeting_id
+      - define_trigger_on:
+          table: nm_committee_manager_ids_user_t
+        relational_column: committee_id
+        log_collection_id:
+          column: user_id
+        log_value:
+          column: committee_id
+      - define_trigger_on:
+          table: user_t
+          update_columns: home_committee_id
+        relational_column: home_committee_id
+        log_value:
+          column: home_committee_id
+        log_collection_id:
+          column: id
   # committee specific permissions
   committee_management_ids:
     type: relation-list
@@ -189,6 +213,14 @@ fields:
       ) AS meeting_ids
     to: meeting/user_ids
     restriction_mode: E
+    log_triggers_data:
+      - define_trigger_on:
+          table: meeting_user_t
+        relational_column: user_id
+        log_collection_id:
+          column: user_id
+        log_value:
+          column: meeting_id
   organization_id:
     type: relation
     reference: organization
--- a/collections/user.yml
+++ b/collections/user.yml
@@ -129,16 +129,13 @@ fields:
       ) AS committee_ids
     log_triggers:
       - on_table: meeting_user_t
-        relational_column: meeting_id
         log_collection_id_column: user_id
         log_value_sql: SELECT committee_id FROM meeting_t WHERE id = ($1).meeting_id
       - on_table: nm_committee_manager_ids_user_t
-        relational_column: committee_id
         log_collection_id_column: user_id
         log_value_column: committee_id
       - on_table: user_t
         on_columns: home_committee_id
-        relational_column: home_committee_id
         log_collection_id_column: id
         log_value_column: home_committee_id
   # committee specific permissions
@@ -205,7 +202,6 @@ fields:
     restriction_mode: E
     log_triggers:
       - on_table: meeting_user_t
-        relational_column: user_id
         log_collection_id_column: user_id
         log_value_column: meeting_id
   organization_id:
--- a/collections/user.yml
+++ b/collections/user.yml
@@ -127,30 +127,20 @@ fields:
           WHERE u_hc.home_committee_id IS NOT NULL AND u_hc.id = u.id
         ) AS ci
       ) AS committee_ids
-    log_triggers_data:
-      - define_trigger_on:
-          table: meeting_user_t
+    log_triggers:
+      - on_table: meeting_user_t
         relational_column: meeting_id
-        log_collection_id:
-          column: user_id
-        log_value:
-          column:
-          sql: SELECT committee_id FROM meeting_t WHERE id = ($1).meeting_id
-      - define_trigger_on:
-          table: nm_committee_manager_ids_user_t
+        log_collection_id_column: user_id
+        log_value_sql: SELECT committee_id FROM meeting_t WHERE id = ($1).meeting_id
+      - on_table: nm_committee_manager_ids_user_t
         relational_column: committee_id
-        log_collection_id:
-          column: user_id
-        log_value:
-          column: committee_id
-      - define_trigger_on:
-          table: user_t
-          update_columns: home_committee_id
+        log_collection_id_column: user_id
+        log_value_column: committee_id
+      - on_table: user_t
+        on_columns: home_committee_id
         relational_column: home_committee_id
-        log_value:
-          column: home_committee_id
-        log_collection_id:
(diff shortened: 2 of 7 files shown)


dev/src/generate_sql_schema.py
```diff
@@ -427,6 +435,32 @@ def get_not_null_trigger_params(type_: str) -> dict[str, str]:
...
+        return {
+            "trigger_type": type_,
+            "changed_item_state": "added" if type_ == "iu" else "deleted",
+            "changed_item_state_phrase": (
+                "added to" if type_ == "iu" else "deleted from"
+            ),
+            "primary_hstore": primary_hstore,
+            "primary_hstore_uppercase": primary_hstore.upper(),
+            "secondary_hstore": "old" if type_ == "iu" else "new",
+            "trigger_return_value": "NEW" if type_ == "iu" else "NULL",
```
peb-adr, line 456: If we use `AFTER` triggers only we wouldn't need this as we would only return `NULL`

dev/src/generate_sql_schema.py
```diff
@@ -2290,6 +2404,75 @@ def get_foreign_key_notify_trigger(
...
+                "log_collection_id_sql": log_collection_id.get("sql") or "",
+                "trigger_column": data["relational_column"],
+                "changed_item_column": log_value.get("column") or "",
+                "changed_item_sql": log_value.get("sql") or "",
+            }
+            subst_iu = {
+                **subst_common,
+                "trigger_type": "iu",
+                "trigger_name": trigger_name_iu,
+                "trigger_operations": f"BEFORE INSERT{trigger_columns_iu}",
```
peb-adr, line 2329: Should always be `AFTER`.
As all changes should only be written to log tables _after_ being executed.

In practice this probably wouldn't be a real difference as all changes are only visible when the transaction is committed.
But I argue for correct semantics (and being safe).

collections/committee.yml
```diff
@@ -52,9 +52,33 @@ fields:
           UNION
 
           -- Select user_id from home committees
-          SELECT u.id FROM user_t u WHERE u.home_committee_id = c.id
+          SELECT u.id FROM user_t u
```
peb-adr: Going for uniformity `FROM` should go on a new line as well.

collections/committee.yml
```diff
@@ -52,9 +52,33 @@ fields:
...
+          WHERE u.home_committee_id = c.id
         ) _
       ) AS user_ids
+    log_triggers_data:
```
peb-adr: Maybe just call it `log_triggers`.

Also: When defining a new field or property the README.md should be adjusted as well.

collections/user.yml
```diff
@@ -189,6 +213,14 @@ fields:
...
     to: meeting/user_ids
     restriction_mode: E
+    log_triggers_data:
+      - define_trigger_on:
+          table: meeting_user_t
+        relational_column: user_id
+        log_collection_id:
+          column: user_id
+        log_value:
+          column: meeting_id
```
peb-adr: We would like to see this hierarchy be flattened.
It might look something like this:

```suggestion
    log_triggers_data:
      - on_table: meeting_user_t
        relational_column: user_id
        log_collection_id_column: user_id  # only one allowed
        log_collection_id_sql: user_id     # only one allowed
        log_value_column: meeting_id
```

hjanott: I understood now that there may be an `update_column` for `trigger_on`. So I would suggest the following instead.
```suggestion
    log_triggers:
      - trigger_on:
          table: meeting_user_t
 #         update_column: {XXX} Optional
        relational_column: user_id
        log_collection_id_column: user_id  # only one allowed
        log_collection_id_sql: user_id     # only one allowed
        log_value_column: meeting_id     
```

dev/src/generate_sql_schema.py
vkrasnovyd: > Should always be AFTER.

Not for this trigger. I used BEFORE because I need to read the old state of the other instance (column of the view) in order to perform check `IF old_log_field_value IS NULL OR NOT (added_item = ANY(old_log_field_value))`.

The value of the calculated field gets re-calculated when the operation is performed. Therefore in the before-trigger I can read the old state of the calculated field for which the log entry has to be written and in the after trigger - the new state. For this reason ud is written as the after-trigger and iu - as before-trigger.

collections/user.yml
hjanott: Or if you prefer `on_table` and `on_update_column` properties. We could do that also.

collections/user.yml
```diff
@@ -127,6 +127,30 @@ fields:
...
         ) AS ci
       ) AS committee_ids
+    log_triggers_data:
+      - define_trigger_on:
+          table: meeting_user_t
+        relational_column: meeting_id
+        log_collection_id:
+          column: user_id
+        log_value:
+          column:
```
hjanott: I don't really like this empty value.

vkrasnovyd: Leftovers from the cleanup. Unintentionally skipped it while removing the other empty values. Will delete.

dev/src/generate_sql_schema.py
```diff
@@ -1939,6 +1985,74 @@ class Helper:
         $check_equals_meeting_id_for_meeting$ LANGUAGE plpgsql;
 
         """)
+    LOG_CALCULATED_ID_ARRAY_TRIGGER_FUNCTION_TEMPLATE = string.Template(dedent("""
+            CREATE OR REPLACE FUNCTION ${trigger_type}_log_modified_calculated_id_array_field()
```
```suggestion
            CREATE OR REPLACE FUNCTION log_${trigger_type}_modified_calculated_id_array_field()
```

vkrasnovyd: Did you mean this?

```suggestion
            CREATE OR REPLACE FUNCTION log_${trigger_type}_modified_calculated_id_array_field()
```

dev/src/generate_sql_schema.py
```diff
@@ -1939,6 +1985,74 @@ class Helper:
...
+            DECLARE
+                log_collection TEXT := TG_ARGV[0];
+                log_collection_id_column TEXT := TG_ARGV[1];
+                log_collection_id_sql TEXT := TG_ARGV[2];
+                log_field TEXT := TG_ARGV[3];
+                trigger_column TEXT := TG_ARGV[4];
+                ${changed_item_state}_item_column TEXT := TG_ARGV[5];
+                ${changed_item_state}_item_sql TEXT := TG_ARGV[6];
+
+                ${primary_hstore}_hstore hstore := hstore(${primary_hstore_uppercase});
```
hjanott: I like this usage of the hstore.

dev/src/generate_sql_schema.py
```diff
@@ -170,6 +171,13 @@ def generate_the_code(
         im_table_code = ""
         errors: list[str] = []
 
+        for type_ in ["iu", "ud"]:
+            pre_code += (
+                Helper.LOG_CALCULATED_ID_ARRAY_TRIGGER_FUNCTION_TEMPLATE.substitute(
+                    cls.get_log_calculated_id_array_trigger_params(type_)
+                )
+            )
+        pre_code += Helper.FILE_TEMPLATE_CONSTANT_TRIGGERS
```
hjanott, line 180: Please do me a favour and move all the log functions together into one block.
Maybe even adding a comment delimiting towards the other parts of the pre code.

dev/src/generate_sql_schema.py
```diff
@@ -1939,6 +1985,74 @@ class Helper:
...
+                    ${changed_item_state}_item := ${primary_hstore}_hstore -> ${changed_item_state}_item_column;
+                END IF;
+
+                IF ${changed_item_state}_item IS NULL THEN
+                    RETURN ${trigger_return_value};
+                END IF;
+
+                -- Add log entry only if log_field value actually changes
+                EXECUTE format('SELECT %I from %I where id = %L', log_field, log_collection, log_collection_id) INTO ${secondary_hstore}_log_field_value;
+                IF ${secondary_hstore}_log_field_value IS NULL OR NOT (${changed_item_state}_item = ANY(${secondary_hstore}_log_field_value)) THEN
```
hjanott: Adjust variable name

dev/src/generate_sql_schema.py
```diff
@@ -2290,6 +2404,71 @@ def get_foreign_key_notify_trigger(
...
+            """))
+        processed_tables: dict[str, int] = {}
+        parts: list[str] = []
+
+        subst_base = {
+            "view_name": view_name,
+            "log_field": log_field,
+        }
+
+        for data in triggers_data:
```
hjanott: Adjust variable name

`triggers_data` -> `log_triggers`

`for log_trigger in log_triggers`

collections/committee.yml
```diff
@@ -52,9 +52,25 @@ fields:
...
-          SELECT u.id FROM user_t u WHERE u.home_committee_id = c.id
+          SELECT u.id
+          FROM user_t u
+          WHERE u.home_committee_id = c.id
         ) _
       ) AS user_ids
+    log_triggers:
+      - on_table: meeting_user_t
+        relational_column: meeting_id
+        log_collection_id_sql: SELECT committee_id FROM meeting_t WHERE id = ($1).meeting_id
```
hjanott: check if relational_column is needed.

dev/src/validate.py
```diff
@@ -324,6 +324,9 @@ def check_field(
             if nested and type in ("relation", "relation-list"):
                 valid_attributes.append("enum")
             valid_attributes.extend(("reference", "deferred", "sql"))
+            if field.get("sql"):
+                valid_attributes.append("log_triggers")
+                self.check_log_triggers(collectionfield, field)
```
hjanott, line 329: Also please update this string.

collections/committee.yml
vkrasnovyd: Done.

collections/committee.yml
vkrasnovyd: Changed to log_triggers. Still have to update README.md.

collections/user.yml
vkrasnovyd: Flattened. Looks more readable indeed.

collections/user.yml
vkrasnovyd: Deleted.

collections/committee.yml
vkrasnovyd: Notes for my future self:
1. No, it should be optional. Needed only if `log_collection_id_sql` or `log_value_sql` is specified.
2. Should be a list of table+columns that define intermidiate relations between the table on which trigger is defined and the one storing `log_value` or `log_collection_id`.
3. In collections: `relational_column` -> `intermediate_relations`. Must be a list of items that include 3 values:
  - intermediate_table
  - intermediate_table_columns
  - id_selection_statement
4. In schema the optional parameters `intermediate_table`, `intermediate_table_columns` and `id_selection_statement` should be added in the end of trigger definition and retrieved in loop. For each group of these 3 values the check should be performed that value is not null (insted of the current `relational_column` check). Something like:
```
    EXECUTE format('SELECT %I from %I where id = %s', intermediate_table_columns, intermediate_table, id_selection_statement) INTO intermediate_relation_value USING NEW/OLD;
    IF intermediate_relation_value IS NULL THEN
        RETURN NEW/NULL;
    END IF;
```
5. First line of the comment should be updated.
6. validator has to be adjusted.

dev/src/validate.py
vkrasnovyd: Which part do you want to be changed and in what way?

dev/src/generate_sql_schema.py
hjanott: Yes. It seems I was concurrently changing my comment while you were writing yours.

dev/src/validate.py
hjanott: This is probably outdated.

dev/src/generate_sql_schema.py
vkrasnovyd: Done for both functions and trigger definitions.

dev/src/generate_sql_schema.py
vkrasnovyd: Done.

dev/src/generate_sql_schema.py
vkrasnovyd: Done.

dev/src/generate_sql_schema.py
vkrasnovyd: Done.

dev/src/validate.py
vkrasnovyd: Probably inside of `self.check_log_triggers()`. I made some adjustments there.

dev/src/validate.py
```diff
@@ -611,6 +620,12 @@ def check_log_triggers(self, collectionfield: str, field: dict[str, Any]) -> Non
                         f"For for item {i} of {collectionfield}.log_triggers value in '{base_attr_name}_column' will be ignored because '{base_attr_name}_sql' is defined."
                     )
 
+            for attr in data.keys():
+                if attr not in valid_attributes:
+                    self.errors.append(
+                        f"{base_error_message}: attribute '{attr}' is invalid."
+                    )
```
hjanott: I would honestly expect this validity check to happen first in the outer loop.

vkrasnovyd: I put it in the end of the loop following the structure of `check_field()`: validating allowed fields before the `if attr not in valid_attributes`. And keeping it inside of the inner loop allows to group all the errors for each item separately.

collections/committee.yml
vkrasnovyd: Evaluated this again. For the log trigger such intermediate checks now look like an overkill to me and I don't see a potential usecase for such a check. I will delete `relational_column` from the collection-files and adjust the generator and validator accordingly.

vkrasnovyd: Done.

dev/src/validate.py
```diff
@@ -588,6 +593,9 @@ def check_log_triggers(self, collectionfield: str, field: dict[str, str]) -> Non
         for i in range(len(log_triggers)):
             base_error_message = f"Error in item {i} of {collectionfield}.log_triggers"
             log_trigger = log_triggers[i]
+            if not isinstance(log_trigger, dict):
+                self.errors.append(f"{base_error_message}: must be a dictionary.")
+                continue
             if not (on_table := log_trigger.get("on_table")):
```
hjanott, line 599: Can also be `elif`.

hjanott (approved):
Looks good to me. One suggestion which is minor.

dev/src/validate.py
vkrasnovyd: I prefer to return/continue early instead of nesting the checks.

peb-adr (approved):

PR #483, opened 2026-04-22
Add models_diff.py

peb-adr:
This is not necessarily needed for transitional release IMO
-> not adding `staging-t1` label to keep it simple

As soon as we are doing the real `4.3.0-staging` it will be included

### DIFF
--- a/dev/scripts/models_diff.py
+++ b/dev/scripts/models_diff.py
@@ -127,13 +127,17 @@ def check_field(collection, model_id, field_name):
     field_value_d1 = D1[collection][model_id][field_name]
     field_value_d2 = D2[collection][model_id][field_name]
 
+    field_value_d1_human_readable = ""
+    field_value_d2_human_readable = ""
+    if field_type == 'timestamp':
+        field_value_d1_human_readable = f"  ({str(datetime.fromtimestamp(field_value_d1))})"
 
     is_equal = compare_value(field_type, field_value_d1, field_value_d2)
 
     if not is_equal:
         DIFF += [f"{collection}/{model_id}/{field_name} of type {field_type} differs."]
-        DIFF += [f"  D1: {field_value_d1}"]
-        DIFF += [f"  D2: {field_value_d2}"]
+        DIFF += [f"  D1: {field_value_d1}{field_value_d1_human_readable}"]
+        DIFF += [f"  D2: {field_value_d2}{field_value_d2_human_readable}"]
 
     # Fields of type generic-relation will additionally appear in an expanded form
     if field_type == 'generic-relation':
--- a/dev/scripts/models_diff.py
+++ b/dev/scripts/models_diff.py
@@ -224,6 +224,13 @@ def check_model(collection, model_id):
             info(f"visited field in D2: {collection}/{model_id}/{field_name} (null value)")
             del D2[collection][model_id][field_name]
 
+    remaining_field_names_d2 = list(D2[collection][model_id].keys())
+    # Finally visiting known new fields
+    if collection in ['organization', 'meeting']:
+        if 'time_zone' in remaining_field_names_d2:
+            info(f"visited field in D2: {collection}/{model_id}/time_zone (new field)")
+            del D2[collection][model_id]['time_zone']
+
     # Remove visited model
     if len(D1[collection][model_id]) == 0:
         info(f"fully visited model in D1: {collection}/{model_id}")
--- a/dev/scripts/models_diff.py
+++ b/dev/scripts/models_diff.py
@@ -2,7 +2,7 @@
 
 from json import load
 from yaml import safe_load
-from datetime import datetime,timezone
+from datetime import datetime,timezone,timedelta
 import sys
 
 
@@ -14,25 +14,35 @@ DIFF = []
 
 
 def list_type_is_equal(l1, l2):
-        is_equal = False
+    is_equal = False
 
-        # EXPECTED DIFF: empty list becomes None
-        if type(l1) is list and l2 is None:
-            if len(l1) == 0:
-                is_equal = True
-        # EXPECTED DIFF: non-empty list may be in different order
-        elif type(l1) is list and type(l2) is list:
-            if sorted(l1) == sorted(l2):
-                is_equal = True
-        # TODO: not sure if this is really expected ...
-        #       apparently migration behavior has changed ...
-        #       revisit with more recent everything.json
-        # IF this is indeed correct to compare this way, the sorted equality
-        # above is obsoleted / included by this.
-            elif set(l1).issubset(l2):
-                is_equal = True
+    # EXPECTED DIFF: empty list becomes None
+    if type(l1) is list and l2 is None:
+        if len(l1) == 0:
+            is_equal = True
+    # EXPECTED DIFF: non-empty list may be in different order
+    elif type(l1) is list and type(l2) is list:
+        if sorted(l1) == sorted(l2):
+            is_equal = True
+    # TODO: not sure if this is really expected ...
+    #       apparently migration behavior has changed ...
+    #       revisit with more recent everything.json
+    # IF this is indeed correct to compare this way, the sorted equality
+    # above is obsoleted / included by this.
+        elif set(l1).issubset(l2):
+            is_equal = True
 
-        return is_equal
+    return is_equal
+
+
+def timestamp_is_equal(s1, s2):
+    # EXPECTED DIFF: UNIX timestamps become ISO timestamps
+    # We assume timestamps were stored in UTC before
+    # -> We might need a pre430 migration ensuring this
+    t1 = datetime.fromtimestamp(s1, timezone.utc) - timedelta(hours=0)
+    t2 = datetime.fromisoformat(s2)
+
+    return t1.timestamp() == t2.timestamp()
 
 
 def check_field(collection, model_id, field_name):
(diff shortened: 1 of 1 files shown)


peb-adr:
From my perspective this is now finished.
I am confident that it reliably verifies the integrity of data after migration 100 completed.
With the latest addition of the repeatable `-v` flag the end user can inspect this process in various level of detail and should be able to provide us with valuable information in case the migration will cause problems in cases we could not catch.

dev/scripts/models_diff.py
```diff
@@ -0,0 +1,392 @@
...
+            if set(l1).issubset(l2):
+                is_equal = True
+
+    return is_equal
+
+
+def timestamp_is_equal(s1, s2):
+    # EXPECTED DIFF: UNIX timestamps become ISO timestamps
+    # We assume timestamps were stored in UTC before
+    # -> We might need a pre430 migration ensuring this
```
hjanott: Is that still a thing? I think this line can be deleted.

dev/scripts/models_diff.py
```diff
@@ -0,0 +1,392 @@
...
+    t1 = datetime.fromtimestamp(s1, timezone.utc) - timedelta(hours=0)
```
hjanott: why subtracting the 0 timedelta?

dev/scripts/models_diff.py
```diff
@@ -0,0 +1,392 @@
...
+    return t1.timestamp() == t2.timestamp()
+
+
+def compare_value(type_, value_d1, value_d2):
+
+    # assume not equal until proven otherwise
+    is_equal = False
+
+    # To be very explicit we will now look at all known field types distinctly,
+    # although in most cases '==' equality should suffice.
```
hjanott: For that reason a comparison like
```    if type_ in ('boolean', 'color'):```
would be much more readable.
An approach with match case would also be doable with the union syntax.
```
case (
    'boolean' 
    | 'color'
    | 'decimal(6)'
):
    is_equal = value_d1 == value_d2
```

File: dev/scripts/models_diff.py [hunk 4]
```diff
@@ -0,0 +1,392 @@
...
+    elif type_ == 'string':
+        is_equal = value_d1 == value_d2
+    elif type_ == 'string[]':
+        is_equal = list_type_is_equal(value_d1, value_d2)
+    elif type_ == 'text':
+        is_equal = value_d1 == value_d2
+    elif type_ == 'text[]':
+        is_equal = list_type_is_equal(value_d1, value_d2)
+    elif type_ == 'timestamp':
+        is_equal = timestamp_is_equal(value_d1, value_d2)
```
hjanott: Just to make sure no type was forgotten.
```suggestion
    elif type_ == 'timestamp':
        is_equal = timestamp_is_equal(value_d1, value_d2)
    else:
        raise Exception("Field type does not exist.")
```

dev/scripts/models_diff.py
```diff
@@ -0,0 +1,392 @@
...
+
+    return is_equal
+
+
+def check_field_empty_list(collection, model_id, field_name):
+    global D1, D2, DIFF
```
hjanott: Only D2 is used here.

dev/scripts/models_diff.py
```diff
@@ -0,0 +1,392 @@
...
+    if len(field_value_d2) != 0:
+        log(4, f"- value is no empty list - not visiting")
+        return
+
+    log(4, f"+ visited field in D2: {collection}/{model_id}/{field_name} (empty list)")
+    del D2[collection][model_id][field_name]
+
+
+def check_field_default(collection, model_id, field_name):
+    global D1, D2, DIFF
```
hjanott: `D1` not used here.

dev/scripts/models_diff.py
```diff
@@ -0,0 +1,392 @@
...
+    log(3, f"check_field: {collection}/{model_id}/{field_name} ...")
+
+    field_type = COLLECTIONS[collection][field_name]['type']
+    field_value_d1 = D1[collection][model_id][field_name]
+    field_value_d2 = D2[collection][model_id][field_name]
+
+    field_value_d1_human_readable = ""
+    field_value_d2_human_readable = ""
+    if field_type == 'timestamp':
+        field_value_d1_human_readable = f"  ({str(datetime.fromtimestamp(field_value_d1))})"
```
hjanott: Why not for the d2 field? Would always be empty string like this.

dev/scripts/models_diff.py
```diff
@@ -0,0 +1,392 @@
...
+    del D2[collection][model_id][field_name]
+
+
+def check_model(collection, model_id):
+    global D1, D2, DIFF
+
+    log(2, f"check_model: {collection}/{model_id} ...")
+
+    field_names_d1 = list(D1[collection][model_id].keys())
+    field_names_d2 = list(D2[collection][model_id].keys())
```
hjanott, line 197: I find this too verbose. It should be enough to use only `model_d1` and `model_d2` by obtaining their reference from the dict. Especially since they are accessed many times. (Not for `del` of the model from the collection dict course).

hjanott (changes requested):
Good job. My change requests are mostly concerning style and readability.

dev/scripts/models_diff.py
peb-adr: Just a remnant from times of uncertainty about timezones :P .
Thought id wouldn't hurt to leave in and would make shifting easier if ever needed.

dev/scripts/models_diff.py
peb-adr: Done

dev/scripts/models_diff.py
peb-adr: I wouldn't want the script to end early. If no condition matched the default `False` is returned and the field will be printed in the end.

dev/scripts/models_diff.py
peb-adr: This function exists to handle the specific case of allowing an empty list that exists only in D2.

Citing from this very file.
```
# Fields not present in D1 may appear in D2
# - as None (null value)
[...]
# - with default value
[...]
check_field_default(collection, model_id, field_name)
# - as empty list
[...]
check_field_empty_list(collection, model_id, field_name)
```

dev/scripts/models_diff.py
peb-adr: This function exists to handle the specific case of allowing default values for fields that exist only in D2.

Citing from this very file.
```
# Fields not present in D1 may appear in D2
# - as None (null value)
[...]
# - with default value
[...]
check_field_default(collection, model_id, field_name)
# - as empty list
[...]
check_field_empty_list(collection, model_id, field_name)
```

dev/scripts/models_diff.py
peb-adr: For all types except the timestamp type field values can be compared by human eye at a glance.

For timestamps this is not the case, as they used to be UNIX timestamps (integers) in D1.
Example output (manually changed value in D1 by 10 seconds):

```
All collections, models and fields have been visited and compared.

Found differences.

speaker/1/begin_time of type timestamp differs.
  D1: 1735041483  (2024-12-24 12:58:03)
  D2: 2024-12-24 11:58:13+00:00
```

Yes, the `field_value_d2_human_readable` will always be empty. I included it for uniformity and potential easy expandability.

peb-adr: However I realize now, that not explicitly stating `timezone.UTC` for `fromtimestamp` results in the UTC offset not being printed which still can lead to confusion. I'm changing it so the output will be like this:

```
All collections, models and fields have been visited and compared.

Found differences.

speaker/1/begin_time of type timestamp differs.
  D1: 1735041483  (2024-12-24 11:58:03+00:00)
  D2: 2024-12-24 11:58:13+00:00
```

So it is _actually_ comparable at a glance as intended.

dev/scripts/models_diff.py
hjanott: Yes. That `raise Exception` could be replaced with just error logging.

dev/scripts/models_diff.py
hjanott: I should be a bit more verbose. I think `D1` and `DIFF` do not need to be made global here.

dev/scripts/models_diff.py
hjanott: I also think here the `D1` dict doesn't need to be made global here.

dev/scripts/models_diff.py
peb-adr: This you will have to explain in person I think

dev/scripts/models_diff.py
hjanott: I don't see the use case for `field_value_d2_human_readable`. So removing that would IMHO be better for readability of the code. Otherwise it's just a point for distraction and confusion. But in the end I don't care enough. ;)

hjanott: > So it is _actually_ comparable at a glance as intended.

That's great!

dev/scripts/models_diff.py
peb-adr: Done

dev/scripts/models_diff.py
peb-adr: Used 'in' syntax now.
Done

dev/scripts/models_diff.py
peb-adr: I decided against changing this for reasons of uniformity.
Hoping for an approve nonetheless :pray:

hjanott (approved):

dev/scripts/models_diff.py
peb-adr: Okay, I see there is value to keeping the global statements smaller.
Reduced to only the necessary vars.

dev/scripts/models_diff.py
peb-adr: Same as above

PR #487, opened 2026-04-23
Format yaml

bastianjoel:


dev/Makefile
```diff
@@ -1,14 +1,21 @@
...
-all: pyupgrade black autoflake isort flake8 mypy sqlfluff
+lint: pyupgrade black autoflake isort flake8 mypy sqlfluff
 
 pyupgrade:
 	pyupgrade --py310-plus --exit-zero-even-if-changed $$(find . -name '*.py')
 
 check-pyupgrade:
 	pyupgrade --py310-plus $$(find . -name '*.py')
 
+cleanup-yaml:
```
rrenkert, line 12: I don't like the usage of docker in this target.
At least it leaves an unused image when executing locally.

dev/Makefile
```diff
@@ -1,14 +1,21 @@
 # Commands inside the container
 paths = src/ tests/
 
-all: pyupgrade black autoflake isort flake8 mypy sqlfluff
+lint: pyupgrade black autoflake isort flake8 mypy sqlfluff
```
rrenkert, line 4: Does this renaming have effect on CI somewhere?
If not, we should add a check with linting.

rrenkert (changes requested):

.github/workflows/pick-to-staging.yml
```diff
@@ -41,20 +40,20 @@ jobs:
           fi
 
       - name: Fetch and checkout latest staging branch
-        if: env.skip_pick != 'true'  # Conditional execution
```
peb-adr, line 44: I tend to put two spaces where the line would end if it wasn't for the inline comment.

But you can change it to one idc.

peb-adr (approved):
Formatting looks good and sensible.

I approve although rrenkert has remarks on make targets.

dev/Makefile
```diff
@@ -1,14 +1,21 @@
...
 
 pyupgrade:
 	pyupgrade --py310-plus --exit-zero-even-if-changed $$(find . -name '*.py')
 
 check-pyupgrade:
 	pyupgrade --py310-plus $$(find . -name '*.py')
 
+cleanup-yaml:
+	docker build -t meta-prettier-image -f Dockerfile-prettier .
+	docker run -d --name meta-prettier-container -v ..:/app/ meta-prettier-image
```
peb-adr: You can `run --rm`, obsoleting `docker rm` after `docker stop`.

From `docker run --help`

    --rm                               Automatically remove the container and its associated anonymous volumes when it exits

peb-adr (commented):
Sigh, this comment wasn't saved to the last review..

dev/Makefile
Janmtbehrens: The reason is for prettiers dependence on node, which is otherwise not used in meta.
We could add a target that just assumes the host machine already has node setup and can execute prettier without issue

dev/Makefile
Janmtbehrens: I have not seen "all" being referenced anywhere. It isn't called in workflows, but the composite make targets are. It should probably remain that way, so you can see which of the linters have failed directly looking at the failed workflow step

dev/Makefile
Janmtbehrens: Thats a game changer

.github/workflows/pick-to-staging.yml
hjanott: I slightly prefer two as well.

bastianjoel: This is not something you can configure.

bastianjoel:
@rrenkert Can we merge this?

peb-adr (approved):

PR #488, opened 2026-04-27
Add max_len rule for all trigger names

hjanott:
Also changed trigger names of not null constraints to have that in their name plus one single trigger name for self referencing nm tables.


vkrasnovyd (approved):
Important improvements. Good to have them.

PR #489, opened 2026-04-27
Add prevent_updates trigger

vkrasnovyd:
1. ~~Switched the definition of prevent_updates triggers back to constraint after-triggers with `INITIALLY DEFERRED`. Reason: conflicts with `check_equals` on cascade delete operations (like meeting.delete with different related instances).~~

Updated helper functions for `check_equals()`: removed `ON UPDATE` checks for constant fields in trigger definitions and also for nm and gm tables.

2. Most of the fields that previously had `constant: true` kept this property name and are now protected with the trigger.

3. `constant_legacy: true` is not used for generating the schema. I introduced this property in the collection-files to preserve the original logic when services can define their own constant-logic that is less strict than database-constraint. There are currently 6 fields of this type:
- Less strict constraint is needed to update models on user.merge and internal user.update:
    - chat_message.meeting_user_id
    - assignment_candidate.meeting_user_id:
- This one actually needs discussion. Maybe it does not have to be `constant_legacy` at all. Value in the field can be updated in the external actions (for example, field is present in the schema of speaker.update) and the possibility of update is covered with tests:
    - speaker.point_of_order
- Can be updated by the internal actions + logic will likely change in the new vote-service, so no need to handle these cases now:
    - option.content_object_id
    - poll_candidate.user_id
    - vote.weight
 
4. Added a couple of missed `constant: true` (for `organization_id` and `meeting_id`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,10 +69,10 @@ Each collection-file has the following format:
       fields the value as to be an id of an existing object.
     - The property `equal_fields` describes fields that must have the same value in
       the instance and the related instance.
-    - The property `constant_strict` describes fields values in which can not be
+    - The property `constant` describes fields values in which can not be
       updated once set (protected by batabase constraint).
-    - The property `constant` describes fields that should be protected from being changed
-      inside of the regular extarnal actions (logic should be defined in the services).
+    - The property `constant_legacy` describes fields that should be protected from being
+      changed inside of the regular extarnal actions (logic should be defined in the services).
 - Restriction Mode:
   The field `restriction_mode` is required for every field. It puts the field into a
   restriction group. See https://github.com/OpenSlides/OpenSlides/wiki/Restrictions-Overview
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Each collection-file has the following format:
     - The property `constant` describes fields values in which can not be
       updated once set (protected by batabase constraint).
     - The property `constant_legacy` describes fields that should be protected from being
-      changed inside of the regular extarnal actions (logic should be defined in the services).
+      changed inside of the regular external actions (logic should be defined in the services).
 - Restriction Mode:
   The field `restriction_mode` is required for every field. It puts the field into a
   restriction group. See https://github.com/OpenSlides/OpenSlides/wiki/Restrictions-Overview
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ Each collection-file has the following format:
           - CASCADE: also delete all models in this foreign key
 - JSON Schema Properties:
     - You can add JSON Schema properties to the fields like `enum`, `description`,
-      `items`, `maxLength` and `minimum`
+      `items`, `maxLength`, `minimum` and `maximum`.
 - Additional properties:
     - The property `read_only` describes a field that can not be changed by an action.
     - The property `default` describes the default value that is used for new objects.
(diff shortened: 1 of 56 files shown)


dev/sql/schema_relational.sql
```diff
@@ -291,6 +291,26 @@ BEGIN
...
+DECLARE
+    collection TEXT := TG_ARGV[0];
+    constant_column TEXT := TG_ARGV[1];
+    old_value TEXT;
+    new_value TEXT;
+BEGIN
+    old_value := hstore(OLD) -> constant_column;
+    IF old_value IS NULL THEN
+        RETURN NEW;
+    END IF;
```
hjanott: That essentially means that an unset field can be set with a value. Is that the behaviour we had before? It's probably necessary since we don't always fill all fields with one action, right?

dev/src/helper_get_names.py
```diff
@@ -385,6 +385,12 @@ def get_not_null_1_1_rel_upd_del_trigger_name(
             table_name, column_name
         )
 
+    @staticmethod
+    @max_length
+    def get_constant_field_trigger_name(table_name: str, fname: str) -> str:
+        """gets the name of constant constraint"""
+        return f"constant_{table_name}_{fname}"
```
hjanott:
```suggestion
        return f"tr_constant_{table_name}_{fname}"
```

dev/src/generate_sql_schema.py
```diff
@@ -1006,30 +1030,32 @@ def equal_fields_state_check(
...
+                on_update_fields.append(field_name)
+        return HelperGetNames.get_table_name(table_field.table), on_update_fields
 
     @classmethod
-    def get_event_string(cls, is_update: bool, fields: list[str]) -> str:
-        if is_update:
-            return f"INSERT OR UPDATE OF {', '.join(fields)}"
+    def get_event_string(cls, on_update_fields: list[str]) -> str:
+        if on_update_fields:
+            return f"INSERT OR UPDATE OF {', '.join(on_update_fields)}"
```
hjanott, line 1058: Small change that requires a little more logic but is much better here.

README.md
```diff
@@ -69,6 +69,10 @@ Each collection-file has the following format:
       fields the value as to be an id of an existing object.
     - The property `equal_fields` describes fields that must have the same value in
       the instance and the related instance.
+    - The property `constant` describes fields values in which can not be
+      updated once set (protected by batabase constraint).
+    - The property `constant_legacy` describes fields that should be protected from being
+      changed inside of the regular extarnal actions (logic should be defined in the services).
```
hjanott:
```suggestion
      changed inside of the regular external actions (logic should be defined in the services).
```

hjanott (changes requested):
I appreciate the amount of research this must have entailed.

dev/src/helper_get_names.py
vkrasnovyd: Done.

dev/sql/schema_relational.sql
vkrasnovyd: Good question, actually. There are currently 5 fields with `constant` and without `required`:
- history_entry.original_model_id
- history_position.original_user_id
- meeting.language (has default)
- option.poll_id (is required in backend create action schema)
- poll.global_option_id

The last case seems to be the only case when we really set the value after creating an instance.

vkrasnovyd: After our discussion of the current state I went ahead with the most gentle change in the chosen direction and replacing `constant` with  `constant_legacy` for poll.global_option_id and oposite side `constant_legacy`. Also removed the part marked by you in this comment.
Other changes would require checking and potentially changing the logic so I don't see them as a good part of this PR.

dev/src/generate_sql_schema.py
```diff
@@ -1585,6 +1601,20 @@ class Helper:
...
 
+        CREATE OR REPLACE FUNCTION prevent_updates() RETURNS trigger AS $constant_field_trigger$
+        DECLARE
+            collection TEXT := TG_ARGV[0];
+            constant_column TEXT := TG_ARGV[1];
+            old_value TEXT := hstore(OLD) -> constant_column;
+            new_value TEXT := hstore(NEW) -> constant_column;
+        BEGIN
+            IF old_value IS DISTINCT FROM new_value THEN
+                RAISE EXCEPTION 'Constant value constraint violated for %/%: % can not be updated once set.', collection, NEW.id, constant_column;
```
```suggestion
                RAISE EXCEPTION 'Constant value constraint violated for %/%: % can not be updated.', collection, NEW.id, constant_column;
```

hjanott (approved):
Now it's great.

PR #509, opened 2026-05-11
Mark collections not managed by the backend

bastianjoel:


vkrasnovyd (approved):

ostcar (approved):
I guess, that entry has no technical reason and is just for the documentation?

vkrasnovyd:
> I guess, that entry has no technical reason and is just for the documentation?

It is required to adjust the backend.

bastianjoel:
@ostcar The backend needs this info so it does not attempt to delete such models on its own. 
The alternative would be to return all deleted fqids in the vote service which in my opinion is the worse alternative. 

But I guess it is also nice for documentation.

PR #512, opened 2026-05-11
Check permissions in validate.py

### LINKED ISSUE #511: Check permissions in meta
Since permissions consistency check compares contents only of the files from the meta repository (permissions.yml and collections files), it should be the part of `/meta/dev/src/validate.py` instead of backend: https://github.com/OpenSlides/openslides-backend/blob/24c2444b4302d8720f87f859ae97c26929a9c1a3/cli/generate_permissions.py#L91-L96

Current state: when there are inconsistencies between permissions.yml and group/permissions, there are no notifications from the meta PR cli-s.

vkrasnovyd:
Resolves #511 

Is already relevant because permissions are getting adjusted to the new vote service.

### DIFF
--- a/dev/src/validate.py
+++ b/dev/src/validate.py
@@ -12,6 +12,7 @@ from .helper_get_names import (
     DEFAULT_COLLECTION_META,
     DEFAULT_COLLECTIONS_DIR,
     KEYSEPARATOR,
+    PERMISSIONS_SOURCE,
 )
 
 MAX_FIELD_NAME_LENGTH = 63
@@ -173,6 +174,7 @@ class Checker:
             for attr, values in data.items():
                 if attr in ["unique_together", "unique_together_strict"]:
                     self.check_unique_together(collection, values, attr)
+        self.check_permissions()
 
     def check_field(
         self,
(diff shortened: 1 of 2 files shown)


dev/src/validate.py
```diff
@@ -372,6 +374,38 @@ def check_unique_together(
...
+            collection: str, permissions_dict: dict[str, Any]
+        ) -> None:
+            for key, value in permissions_dict.items():
+                all_permissions.add(f"{collection}.{key}")
+                if isinstance(value, dict):
+                    flatten_permissions(collection, value)
+                elif value is not None:
+                    self.errors.append(
+                        f"Permissions for '{collection}' contain invalid value: {value} (must be a dict)."
+                    )
```
luisa-beerboom: Make the function shorter by moving this helper function out of it.

Could have it return a set and add those to the `all_permissions`-set as the function is called.

dev/src/validate.py
```diff
@@ -372,6 +374,38 @@ def check_unique_together(
...
+
+        if group_permissions != all_permissions:
+            if missing_permissions := str(all_permissions - group_permissions):
+                self.errors.append(
+                    f"Permissions missing in group/permissions: {missing_permissions}."
+                )
+            if additional_permissions := str(group_permissions - all_permissions):
+                self.errors.append(
+                    f"Permissions missing in {os.path.basename(PERMISSIONS_SOURCE)}: {additional_permissions}."
+                )
```
luisa-beerboom, line 410: You could do `all_permissions.symmetric_difference(group_permissions) and throw a conjoined error.
That way you wouldn't even have to turn `group_permissions` into a set.
It _would_ mean that you can't use separate error messages anymore, so it's more a thing of preference, though.

luisa-beerboom (changes requested):

dev/src/validate.py
vkrasnovyd: In this check I prefer to have separate messages to make it easier for the developer to find the source of inconsistency.

dev/src/validate.py
vkrasnovyd: Done.

dev/src/validate.py
```diff
@@ -377,34 +377,38 @@ def check_unique_together(
...
-            if missing_permissions := str(all_permissions - group_permissions):
+            if missing_permissions := (all_permissions - group_permissions):
                 self.errors.append(
-                    f"Permissions missing in group/permissions: {missing_permissions}."
+                    f"Permissions missing in group/permissions: {', '.join(list(missing_permissions))}."
                 )
-            if additional_permissions := str(group_permissions - all_permissions):
+            if additional_permissions := (group_permissions - all_permissions):
+                self.errors.append(
+                    f"Permissions missing in {os.path.basename(PERMISSIONS_SOURCE)}: {', '.join(list(additional_permissions))}."
```
luisa-beerboom, line 396: Wouldn't it be fine to just use `"{os.path.basename(PERMISSIONS_SOURCE)}: {additional_permissions}."`? It's not as if it would be harder to read...

luisa-beerboom (approved):
One minor remark, you could change if you want to, but I don't really mind either way...

dev/src/validate.py
vkrasnovyd: That's exactly what I was using before commit [a169fd9](https://github.com/OpenSlides/openslides-meta/pull/512/commits/a169fd9320dbcf3551f70bd9a662340731655e2a). :)

I updated it because with join() it better matches the style of the other error message with multiple values in the same file (in check_unique_together).

PR #535, opened 2026-06-17
Add max_yes_vote_amount to polls

potatojuicemachine:
This PR adds the max_yes_vote_amount to the polls as  preparation for limiting the amount of yes votes as seen in this PR: https://github.com/OpenSlides/OpenSlides/pull/7134


ostcar:
The current vote-service is depricated. See https://github.com/OpenSlides/openslides-vote-service/pull/548#issuecomment-4743024828

PR #539, opened 2026-06-27
Check reference in validator

### LINKED ISSUE #415: Don't generate 1r:1r relations
Should be `1r:1t`.

Example in the schema (found only one, got fixed in #407):
```
FIELD 1r:1r => option/used_as_global_option_in_poll_id:-> poll/global_option_id
...
FIELD 1r:1r => poll/global_option_id:-> option/used_as_global_option_in_poll_id
```

Occurs when both sides of 1-to-1 relation have 'reference' attribute. Results in both fields being generated in the corresponding tables instead of the back relation being generated only in the view.

When both sides of 1-to-1 relation contain 'reference' attribute, error should be raised on attempt to generate relational schema, generation should fail.

vkrasnovyd:
Resolves https://github.com/OpenSlides/openslides-meta/issues/415 and makes validator fail when `reference` attribute (or it absence) can cause inconsistencies in schema.

Allowed combinations that result in correct schema:
* relation (r) - relation (x)
* generic-relation (r) - relation (x)
* relation (r) - relation-list (r | x) -> reference for relation-list is ignored, generated schema is the same
* generic-relation (r) - relation-list (r | x) -> reference for relation-list is ignored, generated schema is the same
* relation-list (x) - relation-list (x)
* generic-relation-list (x) - relation-list (x)

### DIFF
--- a/dev/src/validate.py
+++ b/dev/src/validate.py
@@ -324,9 +324,11 @@ class Checker:
                     f"invalid value for 'on_delete' for {collectionfield}"
                 )
             valid_attributes.append("equal_fields")
+            if type != "generic-relation-list":
+                valid_attributes.append("reference")
             if nested and type in ("relation", "relation-list"):
                 valid_attributes.append("enum")
-            valid_attributes.extend(("reference", "deferred", "sql"))
+            valid_attributes.extend(("deferred", "sql"))
             if field.get("sql"):
                 valid_attributes.append("log_triggers")
                 self.check_log_triggers(collectionfield, field)
@@ -586,15 +588,10 @@ class Checker:
             to_collectionfield,
             to_field,
         )
-        if all(
-            [
-                "reference" in field and field["type"] == "relation"
-                for field in [to_field, from_field]
-            ]
+        if reference_error := self.check_reference(
+            from_collectionfield, from_field, to_collectionfield, to_field
         ):
-            self.errors.append(
-                f"The relation fields {from_collectionfield} and {to_collectionfield} both have reference set."
-            )
+            self.errors.append(reference_error)
 
         to = to_field["to"]
         if isinstance(to, str):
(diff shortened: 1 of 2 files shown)


dev/src/validate.py
```diff
@@ -324,9 +324,11 @@ def check_field(
                     f"invalid value for 'on_delete' for {collectionfield}"
                 )
             valid_attributes.append("equal_fields")
+            if type != "generic-relation-list":
+                valid_attributes.append("reference")
```
luisa-beerboom, line 328: What would a reference look like for a `generic-relation` then?

dev/src/validate.py
```diff
@@ -705,6 +702,53 @@ def split_collectionfield(self, collectionfield: str) -> tuple[str, str]:
...
+            return f"One of the relation fields {from_collectionfield} and {to_collectionfield} must have 'reference' set."
+
+        allowed, required = self.get_reference_allowed_required(from_field, to_field)
+        if not allowed and from_field.get("reference"):
+            return f"Relational collectionfield {from_collectionfield} can not have 'reference' attribute."
+
+        if required and not from_field.get("reference"):
+            return f"Relational collectionfield {from_collectionfield} must have 'reference' attribute."
+
+        return None
```
luisa-beerboom: What if both sides are required?

dev/src/validate.py
```diff
@@ -705,6 +702,53 @@ def split_collectionfield(self, collectionfield: str) -> tuple[str, str]:
...
+        if from_field["type"] == "relation-list" and to_field["type"] in [
+            "relation",
+            "generic-relation",
+        ]:
+            return (True, False)
+        if (from_field["type"], to_field["type"]) in (
+            ("generic-relation", "relation-list"),
+            ("generic-relation", "relation"),
+            ("relation", "relation-list"),
+        ):
```
luisa-beerboom, line 799: What about the reverse order (i.e. `("relation", "generic-relation")` and so on)?

vkrasnovyd:
Working notes:

## Correct schema is produced

* relation (r) - relation (x)
* generic-relation (r) - relation (x)
* relation (r) - relation-list (r | x) -> reference for relation-list is ignored, generated schema is the same
* generic-relation (r) - relation-list (r | x) -> reference for relation-list is ignored, generated schema is the same
* relation-list (x) - relation-list (x)
* generic-relation-list (x) - relation-list (x)

## Implemented in schema generator, but generate invalid schema

These relations are implemented in schema generator (before this PR), but the produced schema is not correct. Should become not allowed in both collections validator and schema generator.

Field generated in both tables instead of table-view:
* relation (r) - relation (r)
* generic-relation (r) - relation (r)

2 nm-tables get generated:
* relation-list (r) - relation-list (r) - without required fields

## Not implemented in schema generator

No fields get generated, error is added to the end of the generated schema file. Should also make validator scream.

* relation - generic-relation-list
* generic-relation - generic-relation
* generic-relation - generic-relation-list
* generic-relation-list - generic-relation-list

* relation (x) - relation-list (x)
* relation (x) - relation-list (r)
* generic-relation (x) - relation (x)
* generic-relation (x) - relation-list (x)
* generic-relation (x) - relation-list (r)
* relation-list (r) - relation-list (r) - when any of the sides is required
* relation-list (r) - relation-list (x) (reference on any of the sides) - when any of the sides is required
* generic-relation-list (r) - relation-list (r)
* generic-relation-list (x) - relation-list (r)
* generic-relation-list (r) - relation-list (x)

dev/src/validate.py
vkrasnovyd: As discussed today, sent the working notes below.

dev/src/validate.py
vkrasnovyd: If it is a request for an example, here's one, in agenda_item.yml:
```
  content_object_id:
    type: generic-relation
    reference:
    - motion
    - motion_block
    - assignment
    - topic
    to:
      collections:
      - motion
      - motion_block
      - assignment
      - topic
      field: agenda_item_id
    required: true
    equal_fields: meeting_id
    restriction_mode: A
    constant: true
```

It's a list of collections for all of them.

vkrasnovyd: Or what you mean is that I schould also check the content of the `reference`: that it's a single valid collection for `relation` and `relation-list` or a list of valid collections for `generic-relation`?

dev/src/validate.py
vkrasnovyd: Result is the same: the fields are being generated in both tables instead of table + view. `required` in this case changes only whether the NOT NULL constraint exists.

Similar for the other field combinations.

vkrasnovyd: But in this case `required` refers not to the field attribute but to whether the `reference` attribute is required for this side of the relation. If it is required but not set, the error will be added to self.errors.

dev/src/validate.py
luisa-beerboom: I was just not aware this was possible.

dev/src/validate.py
luisa-beerboom: ah, then the variable should probably get a more verbose name.

luisa-beerboom (approved):
I have one minor thing, then it's good

dev/src/validate.py
vkrasnovyd: Done.

dev/src/validate.py
vkrasnovyd: I actully find it a good idea to validate also the value of reference. I've added these checks. For generic relations checks for invalid/extra values happen in a separate method, so we get only 1 error of this type per field.

luisa-beerboom (approved):

PR #543, opened 2026-07-03
[vote] Move user information from poll_ballot to separat collection

ostcar:
@bastianjoel we (@rrenkert @MSoeb @bspekker ) talked, that it would be nice to save, which delegate send the vote.

We do this, but this information gets lost, when the poll gets anonymized. Currently, we only save the user ids in the field `poll/voted_ids`.

The Idea of this PR is, that we use different collections for the ballot and the user information. On non anonymed polls, its easy to connect them both with a relation.

On anonymized, the collection `poll_ballot_user` does not get deleted, but only the reference to `poll_ballot`.

On secret polls, I already delete all `poll_ballot` objects and recreate them. With this model-change, this has to be done, for any anonymize, so it is not possible to guess the relation between `poll_ballot` and `poll_ballot_user` with there incremented ids.

What do you think about this change? Its a bit late in the process. Would it still be possible for you do implement it in the client?


bastianjoel:
I don't see this getting used currently in the new vote implementation of the client. Therefore I guess its okay for me.

bastianjoel:
Nevermind - with the changes in `poll_ballot` this affects the current client implementation. IMHO the changes are okay, but when you have all necessary changes ready please wait with merging until I adjusted the client. 
We are now at a point where I want to keep the feature branch functional at all times.

ostcar:
The changes are ready:

https://github.com/OpenSlides/openslides-go/pull/289
https://github.com/OpenSlides/openslides-vote-service/pull/555
https://github.com/OpenSlides/openslides-autoupdate-service/pull/1443

vkrasnovyd:
These changes also have to be implemented in the backend. I'll attach PR once it is ready.

collections/poll_ballot_user.yml
```diff
@@ -0,0 +1,33 @@
...
+    type: relation
+    reference: poll
+    to: poll/ballot_user_ids
+    required: true
+    restriction_mode: A
+    constant: true
+  poll_ballot_id:
+    type: relation
+    to: poll_ballot/poll_ballot_user_id
+    restriction_mode: A
```
vkrasnovyd, line 20: How about adding `equal_fields: poll_id`?

ostcar:
@vkrasnovyd I am just curious: Is there a reason you removed the managed_by attribute? Was it only for meeting_delete?

vkrasnovyd:
> @vkrasnovyd I am just curious: Is there a reason you removed the managed_by attribute? Was it only for meeting_delete?

Yes.

vkrasnovyd (approved):

PR #546, opened 2026-07-09
[vote] Write changed user fields in history_entry

vkrasnovyd:


luisa-beerboom (approved):
Reviewed only the last commit, that one looks okay.
I'm not quite so sure about the commits that came in due to the main merge though. Why is that even happening here?

vkrasnovyd:
> Reviewed only the last commit, that one looks okay. I'm not quite so sure about the commits that came in due to the main merge though. Why is that even happening here?

Hmm, didn't do that. Could be because of force push of feature/vote that I see in the timeline here. I'll update the PR to clean that up.

PR #547, opened 2026-07-09
[vote] Support multiple delegations

ostcar:
@hjanott and @vkrasnovyd I try to apply this changes. But `make generate-relational-schema` returns with the message:

```
python -m src.generate_sql_schema
Models file /home/ossi/src/openslides-go/meta/dev/sql/schema_relational.sql created with 1 errors/warnings

    meeting_user/vote_delegations_from_ids: Type combination not implemented: nt:nr on field meeting_user/vote_delegations_from_ids
```

I am not sure how to solve this. I tried to add a `reference` to `vote_delegations_from_ids`, but this did not help.

Is this easy to fix? We would like to add the database change to the new vote service. But the client will not use it anyway. If it is not possible at the moment, then this is not a blocker, but something, we should be able to do in the near future.


vkrasnovyd:
@ostcar you should remove 'reference' on both sides.

List of allowed relations in case you need to make further changes in the future: https://github.com/OpenSlides/openslides-meta/pull/539#issue-4757933815

ostcar:
> @ostcar you should remove 'reference' on both sides.
> 
> List of allowed relations in case you need to make further changes in the future: [#539 (comment)](https://github.com/OpenSlides/openslides-meta/pull/539#issue-4757933815)

Thank you. That helped. I still don't know when to add the reference and when not :)

Could you prepare a PR on the backend? For the moment, the client will never add more then one delegation. So it would be ok for the backend to reject multiple delegations. We want to add this change to the new vote service, that we don't need to change the database, if we add this feature later.

ostcar:
os-go: https://github.com/OpenSlides/openslides-go/pull/299
vote: https://github.com/OpenSlides/openslides-vote-service/pull/566
autoupdate: https://github.com/OpenSlides/openslides-autoupdate-service/pull/1450

@bastianjoel Do you need to update the projector-service?

vkrasnovyd:
@ostcar what should happen to `vote_delegated_to_ids` on user.merge?

Will it fit into the new logic if the resulting user has `vote_delegated_to_ids` that is the result of merging values from this field of all the origin users into a single list?

Current behaviour for the context: `vote_delegated_to_id` is set to the value from the highest ranked model that has the field.

ostcar:
I would say, you should keep the old behaviour. Especially when `users_vote_delegations_max_amount` is set to 1 or a lower value then both users combined, then it is not possible to add both lists together. But we should use an consistent method.

@rrenkert what do you think?

vkrasnovyd:
> I would say, you should keep the old behaviour. Especially when `users_vote_delegations_max_amount` is set to 1 or a lower value then both users combined, then it is not possible to add both lists together. But we should use an consistent method.
> 
> @rrenkert what do you think?

Discussed with @rrenkert. For the initial implementation we'll keep the original behaviour (the option you've described).

vkrasnovyd:
backend is ready: https://github.com/OpenSlides/openslides-backend/pull/3683

vkrasnovyd (approved):

bastianjoel (approved):

PR #551, opened 2026-07-15
[vote] Add field for entitled_users

ostcar:
@vkrasnovyd I would say, we need no migration for that. This field is different then the old [`poll/entitled_users_at_stop`](https://github.com/OpenSlides/openslides-meta/blob/31a750af584bb4c845ab005fba65e56681c58265/collections/poll.yml#L93-L95). We don't have the data in the old system, to create a real `entitled_meeting_users` field. We could only create data, that contains the meeting_users at stop. But this would be wrong.

The client has to work with polls, that do not have the data anyway (when delegation is activated).


ostcar:
I merged the [split poll_ballot collection](https://github.com/OpenSlides/openslides-meta/pull/543) into this PR, since this would contain a lot of merge conflicts in the vote service

vkrasnovyd (approved):
New fields looks good to me.

Data miration indeed seems to be not needed, so I wouldn't mention this field in the Migration.md too.

ostcar:
I think we can merge this now. It only adds fields, that do not harm.

@bastianjoel If its ok for you, you can just click on merge

bastianjoel:
> I would say, we need no migration for that.

I am not sure about that as this is used on the poll detail view. Maybe discuss with @emanuelschuetze.

ostcar:
I created this issue for the question: https://github.com/OpenSlides/openslides-meta/issues/573

Could we merge this PR now? I don't think we have to wait for the answer of the migration question, to merge it.

ostcar:
This PR kind of blocks me. I merge it for now. The merge is only in a feature branch, so there is no harm.

PR #553, opened 2026-07-20
[dammi] Add class variables for external access per collection

hjanott:
This is needed for https://github.com/OpenSlides/openslides-backend/pull/3679.

### DIFF
--- a/dev/src/generate_sql_schema.py
+++ b/dev/src/generate_sql_schema.py
@@ -259,12 +259,11 @@ class GenerateCodeBlocks:
                 table_name_code += code + "\n"
             if code := schema_zone_texts["undecided"]:
                 table_name_code += Helper.get_undecided_all(collection_name, code)
-            code = Helper.get_view_head(collection_name)
-            code += Helper.get_view_body_end(
+            cls.view_sql[collection_name] = Helper.get_view_head(collection_name)
+            cls.view_sql[collection_name] += Helper.get_view_body_end(
                 collection_name, schema_zone_texts.get("view", "")
             )
-            cls.view_sql[collection_name] = code
-            view_name_code += code
+            view_name_code += cls.view_sql[collection_name]
 
             if code := schema_zone_texts["post_view"]:
                 cls.view_sql[collection_name] += code
--- a/dev/src/generate_sql_schema.py
+++ b/dev/src/generate_sql_schema.py
@@ -77,11 +77,12 @@ class SubstDict(TypedDict, total=False):
 
 class GenerateCodeBlocks:
     """Main work is done here by recursing the models and their fields and determine the method to use"""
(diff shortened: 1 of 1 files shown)


dev/src/generate_sql_schema.py
```diff
@@ -242,44 +249,61 @@ def generate_the_code(
...
-            view_name_code += Helper.get_view_head(table_name)
-            view_name_code += Helper.get_view_body_end(
-                table_name, schema_zone_texts.get("view", "")
+                table_name_code += Helper.get_undecided_all(collection_name, code)
+            code = Helper.get_view_head(collection_name)
+            code += Helper.get_view_body_end(
+                collection_name, schema_zone_texts.get("view", "")
             )
+            cls.view_sql[collection_name] = code
+            view_name_code += code
```
vkrasnovyd: How about doing the same as for table_sql?


```suggestion
            cls.view_sql[collection_name] = Helper.get_view_head(collection_name)
            cls.view_sql[collection_name] += Helper.get_view_body_end(
                collection_name, schema_zone_texts.get("view", "")
            )
            view_name_code += cls.view_sql[collection_name]
```

dev/src/generate_sql_schema.py
```diff
@@ -242,44 +249,61 @@ def generate_the_code(
...
+
             if code := schema_zone_texts["post_view"]:
+                cls.view_sql[collection_name] += code
                 view_name_code += code
```
vkrasnovyd: Or even firstly updating cls.view_sql[collection_name] with schema_zone_texts["post_view"] and only then adding its content to view_name_code.

dev/src/generate_sql_schema.py
```diff
@@ -78,6 +78,11 @@ class SubstDict(TypedDict, total=False):
 class GenerateCodeBlocks:
     """Main work is done here by recursing the models and their fields and determine the method to use"""
 
+    table_sql: dict[str, str] = {}
+    view_sql: dict[str, str] = {}
+    alter_table_final_sql: dict[str, str] = {}
+    trigger_sql: dict[str, str] = defaultdict(str)
+    intermediate_sql: dict[str, str] = defaultdict(str)
```
vkrasnovyd: This variable gets updated only once per collection. Can be a simple dict with direct assignment of the value.

vkrasnovyd (changes requested):

dev/src/generate_sql_schema.py
hjanott: Done.

dev/src/generate_sql_schema.py
hjanott: Done.

vkrasnovyd (approved):

PR #555, opened 2026-07-22
[dammi] Remove deferred

vkrasnovyd:


hjanott (approved):
That's likely the right thing to do. 👍

hjanott (approved):
Happily approved.

PR #557, opened 2026-07-23
[dammi] Handle remove

vkrasnovyd:
- Added a bunch of methods for generating statements for migration framework (used in the related backend PR)
- Moved some logic to the new helper methods (get_log_calculated_id_array_trigger_data, split_unique_together_fields, get_varchar_max_length) to use the same calculations results in the migration framework


dev/src/generate_sql_schema.py
```diff
@@ -2257,6 +2252,77 @@ def get_enum_types_definitions() -> str:
...
+    ) -> str:
+        return Helper.get_alter_table_statement(
+            collection_or_table_name, f"DROP CONSTRAINT IF EXISTS {constraint_name}"
+        )
+
+    @staticmethod
+    def get_drop_trigger_statement(
+        collection_or_table_name: str, trigger_name: str
+    ) -> str:
+        return f"DROP TRIGGER IF EXISTS {trigger_name} ON {HelperGetNames.get_table_name(collection_or_table_name)};\n"
```
hjanott: I would like this to crash if a wrong name was calculated. Instead of `IF EXISTS` we could check if the trigger name was used for a drop statement before and then silently skip it for further cases.
```suggestion
    if trigger_name in dropped_triggers:
        return ""
    else:
        dropped_triggers.add(trigger_name)
        return f"DROP TRIGGER {trigger_name} ON {HelperGetNames.get_table_name(collection_or_table_name)};\n"
```

dev/src/generate_sql_schema.py
```diff
@@ -2257,6 +2252,77 @@ def get_enum_types_definitions() -> str:
...
+        return Helper.get_alter_column_statement(
+            collection_or_table_name, column_name, f"DROP {attribute}"
+        )
+
+    @staticmethod
+    def get_drop_table_constraint_statement(
+        collection_or_table_name: str, constraint_name: str
+    ) -> str:
+        return Helper.get_alter_table_statement(
+            collection_or_table_name, f"DROP CONSTRAINT IF EXISTS {constraint_name}"
```
hjanott: Same as for the triggers I guess.

dev/src/generate_sql_schema.py
```diff
@@ -2257,6 +2252,77 @@ def get_enum_types_definitions() -> str:
...
+            HelperGetNames.get_enum_name_for_column(collection_name, column_name)
+        )
+
+    @staticmethod
+    def get_drop_table_statement(collection_or_table_name: str) -> str:
+        return f"DROP TABLE {HelperGetNames.get_table_name(collection_or_table_name)} CASCADE;\n"
+
+    @staticmethod
+    def get_drop_view_statement(collection_name: str) -> str:
+        return f'DROP VIEW IF EXISTS "{collection_name}";\n'
```
hjanott: This will be unused.

hjanott (changes requested):
Nice changes already.

dev/src/generate_sql_schema.py
vkrasnovyd: Deleted.

dev/src/generate_sql_schema.py
vkrasnovyd: I prefer checks like `if trigger_name in dropped_triggers` to happen in the diff generator and not here. Removed `IF EXISTS` from here and will implement the checks in the migration framework.

dev/src/generate_sql_schema.py
vkrasnovyd: Same as for the triggers.

dev/src/generate_sql_schema.py
vkrasnovyd: As discussed, found a usage of it in your PR and returned, but without `IF EXISTS`.

hjanott (approved):

dev/src/helper_get_names.py
```diff
@@ -57,7 +57,6 @@ def __init__(
         ref_column: str = "id",
     ):
         self.table = table
-        self.view = table.rstrip("_t")
```
vkrasnovyd, line 60: 1. Was not used in the code.
2. `TableFieldType.__init__()` is always called with collection name and not with table name - no need to trim (although name `table` is not precise in this case).
3. rstrip trims not '_t' but any symbols present in '_t'. Example: when self.table is `assignment`, self.view = `assignmen`.

hjanott: Yeah. I would like to see a PR in the future cleaning up this name mess. How about we do that at a different time? Places like line 999 in generate_schema_helper could make use of an easy to get table name.
If you still want to delete it, lets at least put a warning into here.

vkrasnovyd: I'd prefer it to be deleted already as these names are too confusing and there's no place in the code where we would want to use the current implementation of `TableFieldType.view`. I've found this line when I tried using this attribute in the backend PR which led to errors and incorrect intermidiate results.

Added TODOs in that place for the future cleanup.

hjanott (approved):

dev/src/helper_get_names.py
hjanott: > and there's no place in the code where we would want to use the current implementation of `TableFieldType.view`.

When we do the cleanup, we would want to use it in the line 1053 of `generate_schema_helper` for instance. Using the real table name would lead to changing gm table column names. However, the current implementation is broken.

dev/src/generate_schema_helper.py
```diff
@@ -854,6 +855,19 @@ def get_inline_generated_always_as_constraint(
             f"GENERATED ALWAYS AS (CASE WHEN split_part({own_column}, '/', 1) = '{foreign_table}' THEN cast(split_part({own_column}, '/', 2) AS INTEGER) ELSE null END) STORED",
         )
 
+    @staticmethod
+    def split_unique_together_fields(fields: str) -> list[str]:
+        return [field_name.strip() for field_name in fields.split(",")]
+
+    @staticmethod
+    def get_varchar_max_length(fdata: dict[str, Any], type_: str) -> int | None:
```
hjanott:
```suggestion
    def get_varchar_max_length(fdata: dict[str, Any], type_: str | Decimal) -> int | None:
```

hjanott (changes requested):
I wonder why the linters aren't complaining about this.

dev/src/generate_schema_helper.py
vkrasnovyd: > I wonder why the linters aren't complaining about this.

Because the method is never called with anything but string as value for `type_` (field_def["type"]). So there's no mismatch between the type in the function definition and actual value passed to the function.

The check `elif isinstance(type_, Decimal):` is currently always falsy.  Decimal fields are generated wihtout using this method. I'll update it.

vkrasnovyd: Simplified it. `validate.py` checks that `maxLength` is defined only on fields of types "string" or "text" -> method can return only `int`.

hjanott (approved):

hjanott (approved):

PR #562, opened 2026-07-30
[dammi] Equal fields rework

vkrasnovyd:
This PR restructures functions that generate trigger definitions for `check_equals`, `check_equals_multi` and `check_equals_meeting_id_for_meeting` in a way that trigger names and tables they are defined on can be more easily calculated without the context from `generate_sql_schema.py`. The goal is to re-use these methods in the new migration framework.

### DIFF
--- a/dev/src/generate_sql_schema.py
+++ b/dev/src/generate_sql_schema.py
@@ -2294,9 +2294,9 @@ FOR EACH ROW EXECUTE FUNCTION log_modified_related_models('{foreign_table}', '{r
         return "".join(parts)
 
     @staticmethod
-    def get_nm_table_for_n_m_relation_lists(
+    def get_nm_table_name_and_fields(
         own_table_field: TableFieldType, foreign_table_field: TableFieldType
-    ) -> tuple[str, str, str, str]:
+    ) -> tuple[str, str, str]:
         nm_table_name = HelperGetNames.get_nm_table_name(
             own_table_field, foreign_table_field
         )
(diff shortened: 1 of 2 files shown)


dev/src/generate_sql_schema.py
```diff
@@ -718,14 +720,18 @@ def get_relation_list_type(
...
                         text["create_trigger_equal_fields_code"] = (
-                            cls.get_trigger_check_equal_fields_for_n_m(
+                            cls.get_trigger_definitions_check_equals_multi(
+                                equal_fields,
                                 own_table_field,
                                 foreign_table_field,
                                 nm_table_name,
                                 own_intermediate_field,
                                 foreign_intermediate_field,
+                                False,
```
hjanott: For better readability maybe put the keyword.

dev/src/generate_sql_schema.py
```diff
@@ -1086,45 +1086,63 @@ def get_event_string(cls, on_update_fields: list[str]) -> str:
...
         sql = ""
         for equal_field in equal_fields:
+            if specified_relation_field is None:
+                own_column = own_table_field.column
+            else:
+                own_column = specified_relation_field
+                if (
+                    "reference" in own_table_field.field_def
+                    and "reference" in foreign_table_field.field_def
+                    and foreign_table_field.field_def.get("type") in ["relation"]
```
hjanott: Please check for which relations this is relevant. Can be thrown out if checked by validator.

dev/src/generate_sql_schema.py
```diff
@@ -2559,6 +2518,87 @@ def get_trigger_for_generic_intermediate_table(
 """
         return trigger_text
 
+    @classmethod
```
hjanott: Make these two static.

hjanott (changes requested):

dev/src/generate_sql_schema.py
vkrasnovyd: Done.

dev/src/generate_sql_schema.py
vkrasnovyd: Done.

dev/src/generate_sql_schema.py
vkrasnovyd: Checked and moved from `else` to `if`.

This combination of types and `reference` in general is allowed in validator. This check is more strict than the one in the validator so we still need it.

hjanott (approved):
lgtm

PR #563, opened 2026-07-30
[dammi] extract_formatted_default_value

hjanott:


dev/src/generate_sql_schema.py
```diff
@@ -2673,18 +2691,9 @@ def get_initials(
...
-            else:
-                raise Exception(
-                    f"{table_name}.{fname}: seems to be an invalid default value"
-                )
+            default_value = Helper.get_formatted_default_value(
+                table_name, fname, default, type_
+            )
             subst["default"] = Helper.get_inline_default_constraint(
                 table_name, fname, default_value
             )
```
vkrasnovyd, line 2699: `default_value` is used only once. With the new helper method you can calculate it directly in `get_inline_default_constraint`. Up to you.

vkrasnovyd (approved):
A change that we need.
I have only one suggestion.

dev/src/generate_sql_schema.py
hjanott: Thank you for the suggestion. However, I find this more aesthetically pleasing.

